PR #1, opened 2019-09-04
Autotools.postprocessing

underwoo:
This is the start of a GNU autotools configure build of FRE-NCtools.  What this currently has is something that will build all current tools with both the GNU and Intel compilers.  There are still some things I would like to complete before getting this merged into master:

* [ ] Get `make check` running some tests
* [ ] Update the documentation, in particular the build/install instructions
* [ ] Update the GitLab CI tests.
* [ ] Convert the site-configs directory to config.site files


underwoo:
Already committed.

PR #3, opened 2020-01-13
Search libm for ceil()

marshallward:
On many Linux platforms, libm is not linked by default with libc and
must be explicitly added (-lm).  When building these tools on Hera, libm
was not linked and the ceil() function was unavailable.

This patch adds a search for ceil() in configure.ac, which resolves the
build on Hera.

Note that there are many missing libm functions (e.g. my own laptop
fails because of a missing acos() function) so there might be a more
general way of forcing libm to be used.


underwoo:
This PR has been migrated to our GitLab instance to allow additional testing.  Once the merge is performed in GitLab, the master branch will be synchronized, closing this PR.

PR #15, opened 2020-10-20
Improvements for CI and workaround for test

rem1776:
Updates the github action to use docker containers to compile and run tests with and without mpi enabled, and will save the log file upon failure.

Also added in a check to skip Test24 since it fails with gcc, and adds Test26 as a workaround for Test24.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -1,24 +1,28 @@
 # Github actions CI file
 # Ryan Mulhall 2020
-# Configures with and without mpi and runs make check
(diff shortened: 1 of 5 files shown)


ceblanton:
Hi @ngs333 this update looks good to me. It resolves the failing fregrid reference test by skipping the existing test (done with icc) unless the compiler is icc, and adds a 2nd fregrid reference test with a nccmp float tolerance option.

ngs333 (approved):

PR #26, opened 2020-12-03
Updates for C99 standards compliance

underwoo:
Updated to c99 standards and resolved several compile time warnings.


tools/libfrencutils/mpp.c
```diff
@@ -130,16 +130,16 @@ void mpp_sync()
...
-#ifdef use_libMPI    
+#ifdef use_libMPI
   MPI_Status status;
   /* make sure only one message from pe->to_pe in queue */
   if(request[to_pe] != MPI_REQUEST_NULL) {
     MPI_Wait( request+to_pe, &status );
   }
-  void* temp_data = data; 
-  MPI_Isend(temp_data, size, MPI_DOUBLE, to_pe, tag, MPI_COMM_WORLD, request+to_pe);
+  /*void* temp_data = data; */
```
ngs333, line 139: Is there a reason for leaving the comment?

tools/make_hgrid/create_lonlat_grid.c
```diff
@@ -9,6 +9,7 @@
 #include "tool_util.h"
 #include "constant.h"
 #include "create_hgrid.h"
+#include "create_xgrid.h"
```
ngs333, line 12: Adding this line does make a huge difference! :-)

ngs333 (approved):

tools/libfrencutils/mosaic_util.c
```diff
@@ -521,12 +521,12 @@ double spherical_angle(const double *v1, const double *v2, const double *v3)
...
   ddd = (px*px+py*py+pz*pz)*(qx*qx+qy*qy+qz*qz);
-  if ( ddd <= 0.0 ) 
+  if ( ddd <= 0.0 )
     angle = 0. ;
   else {
     ddd = (px*qx+py*qy+pz*qz) / sqrt(ddd);
-    if( fabs(ddd-1) < EPSLN30 ) ddd = 1;
-    if( fabs(ddd+1) < EPSLN30 ) ddd = -1;
+    if( fabsl(ddd-1) < EPSLN30 ) ddd = 1;
+    if( fabsl(ddd+1) < EPSLN30 ) ddd = -1;
```
bensonr, line 529: fabsl is valid only when NO_QUAD_PRECISION is false.  Otherwise, we would need to use fabs.  I recommend adding a macro in the #ifdef/#else/#endif at 508ff to choose the correct version of the 'abs' function.

tools/make_hgrid/create_gnomonic_cubic_grid.c
```diff
@@ -541,17 +541,17 @@ void direct_transform(double stretch_factor, int i1, int i2, int j1, int j2, dou
 
   for(j=j1; j<=j2; j++) for(i=i1; i<=i2; i++) {
       l = j*nxp+i;
-      if ( fabs(c2m1) > EPSLN7 ) {
-	sin_lat = sin(lat[l]); 
-	lat_t   = asin( (c2m1+c2p1*sin_lat)/(c2p1+c2m1*sin_lat) );
+      if ( fabsl((long double) c2m1) > EPSLN7 ) {
```
bensonr, line 544: Will this casting to a long double work if the system/compiler does not support them.  I assume in the case of no support, the long double is the same as a double, but not 100% sure.

This comment also applies to the succeeding uses of fabsl with casting in this source file.

tools/libfrencutils/mpp.c
underwoo: No.  It was an oversite.  I'll push an update to remove the comment.

tools/make_hgrid/create_gnomonic_cubic_grid.c
underwoo: Thank you for your review.  I'll correct these areas for no quad precision and push an update.

tools/libfrencutils/mosaic_util.c
underwoo: @bensonr should quad precision be enables on all systems that support `long double`, or should it be something the user turns on/off with an option to `configure`?

bensonr:
@underwoo when it comes to the grid generation tools, I think it best to
have it as an option for the user to configure.  The unexpected change from
system to system can and will impact the users who are creating grids for
any purpose (regridding, new experiments, etc).  However, as this is
determined at build time, users on a shared system will not necessarily
know what was used during the build.  While it would be nice to have this
as a runtime option, I don't know how we'd accomplish that without adding
lines of code.  I think our best course of action is to add an attribute to
the NetCDF file when appropriate.

bensonr:
One other thing I forgot in my review was the use of the _AIX macros.  Those were removed from the FMS project and it's probably safe to do the same here.

tools/libfrencutils/mosaic_util.c
underwoo: Looking over the remaining code, `long double` is used in many places outside of the `NO_QUAD_PRECISION`.  `NO_QUAD_PRECISION` is only used in `mosaic_util.c`  and only in the `spherical_angle` function.  Thus, I don't know if it is worth leaving in this this option, and leave it to the compiler to know if `long double` has more precision than `double`.  From what I can tell in my research, all current C compilers support `long double`, but do not guarantee a larger range or precision over `double`.

I can add a test to `configure.ac` to see if `long double` has more precision than `double` and add a flag `--with-quad-precision` (default to yes) to issue a warning/error if quad-precision does not have any addition precision.  Thoughts?

And as an FYI, I found that the C preprocessor has the ability to do compile time substitution of the correct symbol based on the lvalue.  Thus, the following can choose between the `long double` and `double` form of `abs`:

```
#define fabs(X) _Generic((X), \
     long double: fabsl, \
         default: fabs)(X)
```

Then in the code, you can do `ans = fabs(var)`, and the preprocessor will use the correct form of `fabs` based on the type.

tools/libfrencutils/Makefile.am
```diff
@@ -9,6 +9,7 @@ libfrencutils_a_SOURCES = affinity.c \
                           constant.h \
                           create_xgrid.c \
                           create_xgrid.h \
+                          fms_platform.h \
```
bensonr, line 12: The git comparison shows fms_platform.h has been deleted.

tools/libfrencutils/mosaic_util.c
```diff
@@ -721,14 +736,14 @@ void mult(long double m[], long double v[], long double out_v[]) {
...
+
   const long double det =  m[0] * (m[4]*m[8] - m[5]*m[7])
                      -m[1] * (m[3]*m[8] - m[5]*m[6])
                      +m[2] * (m[3]*m[7] - m[4]*m[6]);
 #ifdef test_invert_matrix_3x3
   printf("det = %Lf\n", det);
-#endif  
-  if (fabs(det) < EPSLN15 ) return 0;
+#endif
+  if (fabsl(det) < EPSLN15 ) return 0;
```
bensonr, line 746: I think this needs to be reverted to fabs to work with the macro inserted into mosaic_util.h

underwoo: `det` is defined a few lines above as `const long double det`.  Since `det` is always `long double` this is the correct abs function.

tools/libfrencutils/Makefile.am
underwoo: I just pushed an update removing this line from the makefile.

PR #28, opened 2020-12-08
Small update to gaea config files

ceblanton:
so that the README.md compile instructions work on gaea


ceblanton:
Hi @ngs333, I think the ncrc compile options may not be ideal, but this at least is getting the ncrc site config files in the right direction

ngs333 (approved):

PR #29, opened 2020-12-09
Re-add hydrology dataset symlinks

ceblanton:
The hydrology grid-generating script (make_simple_hydrog.csh) uses two input files, which are too large to be included in the repo. For now we will symlink to GFDL's filesystem but plan to keep these and other large files in a GDFL-managed git LFS server.

This update lets "make install" install the hydrology script dependencies properly at GFDL

### DIFF
--- a/t/Test25-hydrology.sh
+++ b/t/Test25-hydrology.sh
@@ -1,5 +1,4 @@
 #!/usr/bin/env bats
-set -x
 
 @test "Wrapper complete hydrology test" {
 
--- a/t/Test25-hydrology.sh
+++ b/t/Test25-hydrology.sh
@@ -1,4 +1,5 @@
 #!/usr/bin/env bats
+set -x
 
 @test "Wrapper complete hydrology test" {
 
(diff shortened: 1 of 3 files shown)


ceblanton:
@ngs333 these two symlinks were removed (replaced with empty files) during one of the big merges. I think that was accidental? If not don't merge this in!

ngs333:
@ceblanton Their removal must have been accidental.

ceblanton:
@ngs333 or @underwoo , can you advise on the merits (or lack there of) and path forward for this PR? The motivation is to have `make install` install these large hydrology input files (not in this repo; on GFDL's FMS filesystem as they are too large).

However, creating symlinks that will be broken on all sites except GFDL seems questionable, and indeed cause test failures (that I don't understand, admittedly). git-lfs seems like a solution down the road, but for now, what is the best (or least bad) way to have `make install` install these two files at GFDL and not cause test failures generally?

underwoo:
I think it would be better, at least until we get LFS working as we like, to place the data files on the FTP site (or data portal) with information in the README or other document on how to download the files and install them.  I don't like the idea of a symlink that will only work on some systems.  

In order to run the tests, we can use an environment variable (or configure option -- e.g. --with-datafiles=/location/of/files) to know if the tests can be run.  There are some _fancy_ things we could do depending if the user had downloaded/extracted the files in the source directory.  We can explore these if desired.

underwoo:
@ceblanton is this still needed?  If not, please close.

PR #27, opened 2020-12-04
Grid generation: multiple nest capability

A-Kyle:
This branch includes Bill Ramstrom's (AOML/HRD) original modifications to the make_hgrid code that allow it to generate multiple nested grids on the cubed sphere. The usage message for the program has been updated to reflect changes to its functionality. 

Example command-line statement to generate grids for all 6 tiles of the cubed sphere and 2 nests on tile 6:

`./make_hgrid --grid_type gnomonic_ed --nlon 192 --grid_name C96_grid --do_schmidt --stretch_factor 1. --target_lon -57.0 --target_lat 0.0 --nest_grids 2 --parent_tile 6,6 --refine_ratio 4,4 --istart_nest 81,151 --jstart_nest 81,39 --iend_nest 112,182 --jend_nest 112,70 --halo 3 --great_circle_algorithm`

PR #34, opened 2020-12-16
CI container update and fix for log files

rem1776:
Updates the GitHub actions file to pull and run on a stored container image, and removes the docker recipe file and run script.

The Github Action's steps will run on the container directly as opposed to building an image and then running through docker commands and a run script, which fixes an issue with the testing logs getting saved and allows you to easily see if a build failed from configuring, building, or testing.

***side note: the last merge (with #27) seems to still fail with these updates when i tested the merge. From the log file, it looks like some sort of memory allocation error but I'm not sure why it would only be happening within the CI. You can download it [here](https://github.com/rem1776/FRE-NCtools/actions/runs/426447936).


.github/workflows/main.yml
```diff
@@ -1,28 +1,35 @@
...
-  build:
+  CI:
     runs-on: ubuntu-latest
     strategy:
       matrix:
         with_mpi: ['','--with-mpi']
         enable_quad_precision: ['', '--enable-quad-precision']
-    env:
+    container:
+      image: ryanmulhall/fre-nctools-base
```
ngs333, line 15: Ryan, Would you know if the directory of an individual is required to be used here?

rem1776: Currently we don't have a central public container registry or dockerhub account and to run the action this way requires a registry to pull the image from. I believe this should change soon though as the container project moves along and then i can migrate it to whichever to solution we go with.

ngs333 (approved):

Issue #45, opened 2021-01-26
poly_area() gives wrong values for area when the side of polygon crosses a pole


nikizadehgfdl:
**Describe the bug**
The exchange grid tool uses a function poly_area(), which returns the area of a polygon on globe with given vertices, to calculate the (exchange) grid cell areas. This function gives an unreasonably large values when the side of such polygons crosses a pole. This is because the line inegral formula which the function is based on is not valid when such crossing happens and is bigger than what it should be by pi*R**2 (which explains why the returned area is almost half the area of the Earth).

This pole crossing happens mostly for stretched ATM grids (non-GCA) giving rise to tiling errors for those grids and holes in the Antarctic land_mask.
  
**To Reproduce**
Get a version of frenctools with more verbosity for the exchange grid creation tool:
```
bash
git clone -b nnz_NONGCA_strechedGridIssues https://gitlab.gfdl.noaa.gov/fre/fre-nctools.git fre-nctools_gitlab
cd fre-nctools_gitlab
git checkout adde039  #adds calculated atm_area to land_mask files
export CONFIG_SITE=site-configs/gfdl-ws/config.site 
source site-configs/gfdl-ws/env.sh 
module load gcc/6.2.0
export PATH=.:$PATH
autoreconf -i
./configure 
make
```
Now create a coupled mosaic with a stretched grid for ATM such as c256r25tlat32.0_om4p25.
Explore the values of area_atm array along the longitude in the tile that contains the south pole (land_mask_tile3.nc) . See figure below.

**Expected behavior**
area_atm should not give huge values.

**System Environment**
fre-nctools on PAN 

**Additional context**
See issue #44 for the inaccuracies in poly_area even for non-stretched grids. 

Figure below shows the value of area_atm in land_mask_tile3.nc
```
pathname320='/work/Niki.Zadeh/MOM6-examples_myfork/ice_ocean_SIS2/OM4_025/preprocessing.20201028/mosaic_c256r25tlat32.0_om4p25_githubMaster_Verbose_20210121/'          
land_mask0 = nc.Dataset(pathname320+'land_mask_tile3.nc')
mask320= np.array(land_mask0.variables['mask'])
area_atm320= np.array(land_mask0.variables['area_atm'])
plt.plot(area_atm320[128,:],marker='o')
```

![atm_area_32N](https://user-images.githubusercontent.com/6394244/105796513-f9f2c900-5f5c-11eb-9fd6-19b924a1aa67.png)

[...]

Issue #44, opened 2021-01-26
CubedShere grid cell area calculated by exchange grid tool has discontinuous values around the two poles


nikizadehgfdl:
**Describe the bug**
The cubed sphere grid cell area calculated by the exchange grid tool has bumps around the two poles  (see figures below). These bumps are not there in the cell area calculated by make_hgrid itself, the tool that creates the CS grid. Note that make_hgrid uses spherical_excess_area() (Gauss-Bonnet formula area_of_polygon = sum of interior spherical angles - 2pi) , whereas the exchange grid tool uses poly_area()  (  area_of_polygon = line_intergal sin(lat) d(lon) ). The question is which one of these formulas are more accurate near the singular poles?  

Non-GCA Stretched grids have the same symptoms which results in tiling errors.
  
**To Reproduce**
Get a version of frenctools with more verbosity for the exchange grid creation tool:
```
bash
git clone -b nnz_NONGCA_strechedGridIssues https://gitlab.gfdl.noaa.gov/fre/fre-nctools.git fre-nctools_gitlab
cd fre-nctools_gitlab
git checkout adde039  #adds calculated atm_area to land_mask files
export CONFIG_SITE=site-configs/gfdl-ws/config.site 
source site-configs/gfdl-ws/env.sh 
module load gcc/6.2.0
export PATH=.:$PATH
autoreconf -i
./configure 
make
```
Now create any coupled mosaic grid with a uniform CS grid, e.g., Test03-grid_coupled_model.sh
Plot the calculated area_atm in land_mask_tile6.nc

**Expected behavior**
computed area in C48_grid.tile6.nc does not have a bump at the pole.

**System Environment**
FRE-NCTOOLS build on PAN

**Additional context**
We already know that the poly_area() formula is not correct when side of polygon crosses a pole. 

Figure below shows area of ATM grid cells along the longitude passing through pole (at the peak) calculated by the exchange grid tool (poly_area) for c256_om4p25 grid
![atm_area_spherical_excess_area](https://user-images.githubusercontent.com/6394244/105786920-f4d94e00-5f4b-11eb-8121-9b989a0ffd89.png)

Same ATM grid cell area as above calculated by the make_hgrid.c (spherical_excess_area())
![atm_area_polyarea](https://user-images.githubusercontent.com/6394244/105787053-2a7e3700-5f4c-11eb-9ff5-4a9b493a329f.png)

ngs333:
@nikizadehgfdl 
[...]

PR #60, opened 2021-02-17
Fix poly_area when side of grid cell polygon crosses pole

nikizadehgfdl:
Fix poly_area when polygon side crosses a pole fixes issue #45 
    
    - Added the derivation for poly_area formula to the comments .
    - The area needs a fix for the case that the polygon side crosses a pole (d(longitude)=+- pi).
      In that case the  area formula was missing an additive factor of pi.
    - Missing a pi in the formula results a huge discrepancy of the order of the Earth area!
    - Similarly, if the polygon has a corner with angle A, then the formula
      needs an additive factor of A (in radians). But this latter case is
      already handled by fix_lon() and is the reason it creates a twin pole vertex
      (with longitudes paired with the preceding and following vertices)
      if the original polygon has a vertex at the pole.
    - Put the fix also in fix_lon() too. 
    - Investigate why putting the fix in poly_area alone is not enough?
      If I put the fix in poly_area only some cells near NP come with inaccurate area.
    - Unit tests in create_xgrid.c in PR #59 includes area calculation for these special grid box and show this fix works.
    - This does not fix issue #44  which is due to a numerical inaccuracy that has to be worked out separately (probably via quadrature). 

@ceblanton please assign this for review to @bensonr and @ngs333 or anyone who might be more familiar with poly_area calculation.

### DIFF
--- a/tools/libfrencutils/mosaic_util.c
+++ b/tools/libfrencutils/mosaic_util.c
@@ -407,11 +407,16 @@ double poly_area(const double x[], const double y[], int n)
     /*When the path goes through a Pole the line integral cannot be
       approximated by the fomula below. But we can show for these
       special grid cells the contribution of such paths to the area is pi.
-      We apply a fix to fix_lon function by adding a couple
+      We may also apply a fix to fix_lon function by adding a couple
       of twin poles to the polygon side that crosses the pole (diagnosed by dx=+/-pi).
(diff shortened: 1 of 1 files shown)


bensonr (approved):
Nice work @nikizadehgfdl

ngs333 (approved):

PR #64, opened 2021-03-11
Change container to a new ubuntu focal version to fix CI

rem1776:
Updates the container in which the CI runs to a version Seth had made with a newer version of Ubuntu and nccmp. Fixes the CI crash.


ngs333 (approved):

ceblanton (approved):
Nice! Great that we have a noaagfdl dockerhub account now too

PR #66, opened 2021-03-15
Add prepend-, append- and remove-path subcommands to env.sh scripts

underwoo:
Added additional subcommands to placeholder `module` to allow adding paths to env variables


ceblanton (approved):
Thanks @underwoo the updates look good and work properly on the 3 GFDL sites.

PR #73, opened 2021-03-25
Test.updates

underwoo:
@rem1776 I updated a few of the tests using additional `bats` items.  This should do a better job at cleaning up after the tests (even failed tests).  I needed to do this while tracking down a failing tests -- the old method did not clean up after a test failed, causing future tests to fail.

I suggest updating the other tests with similar modifications.

### DIFF
--- a/t/Test10-remap_land_res.sh
+++ b/t/Test10-remap_land_res.sh
@@ -22,53 +22,42 @@
 
 # Test remap_land: remap land restart files.
 
-@test "Test remap_land: remap land restart files" {
-  if [ ! -d "Test10" ] 
-  then
-  		mkdir Test10
-  fi
-
-  cd Test10
-
-  mkdir src_restart
-  cd src_restart
-  cp $top_srcdir/t/Test10-input/src_restart/* . 
-  files=`ls`
-  for file in $files
+# Prepare takes a directory name, and generates the required netCDF input
+# files for a given test.  prepare_input_data expects a single argument <dir_name>
+# which is the directory name in $top_srcdir/t/Test10-input/<dir_name>
+# that contains the input data
+prepare_input_data ()
+{
+  local inputdir=$top_srcdir/t/Test10-input
+  for dir in $( ls -1 $inputdir )
   do
-	 ncgen -o ${file:0:${#file}-1} ${file}
+    local datadir=$inputdir/$dir
+    mkdir -p $dir
+    # Ensure the directory is empty
+    rm -rf $dir/*
+    # Generate the restart files
+    for file in $( ls -1 $datadir/*.ncl )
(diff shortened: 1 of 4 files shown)


rem1776 (approved):
Sounds good, I'll start looking into the other test files

t/Test12-mppnscatter.sh
```diff
@@ -20,131 +20,42 @@
...
+  prepare_input_data
+  
+  # Scatter the file
+  run command mppncscatter -i 2 -j 3 -x 2 -y 12 input/$test_file
   [ "$status" -eq 0 ]
 
-#Combine the file: 
-  run command mppnccombine -64 fv_core.res.tile6.nc fv_core.res.tile6.nc.???? 
+  # Combine the file:
+  run command mppnccombine -64 output/$test_file ${test_file}.????
```
ceblanton, line 59: Your updates didn't change this, but I wonder why the nccmp's were commented out previously. It would be good to verify that`input/$test_file` is identical to `output/$test_file`. If we wanted to do that, we could remove the `-64` from the `mppnccombine` so that the output file would be the same format as the input file, then add a nccmp check right after:

```
#  Verify that the recombined file is identical to the unscattered file
run command nccmp -d input/$test_file output/$test_file
[ "$status" eq 0 ]

**Review by ceblanton — APPROVED** [review]
Nice! The bats supported mechanisms for the test directory and teardown is much cleaner, consistent, and safer, agreed.

I suggested adding (or uncommenting, really) a nccmp to the scatter test but it was a preexisting condition so won't hold this up for it.

File: t/Test12-mppnscatter.sh (hunk 1 above)
Comment (underwoo) [reply t/Test12-mppnscatter.sh:59]: @ceblanton I think some of the extra commenting out of tests was to get the tests to pass.  I agree we should revamp all the tests to ensure we are testing all the correct things, including additional options, build options, etc.

Please put in an issue to remind us.

Comment (ceblanton) [reply t/Test12-mppnscatter.sh:59]: Good idea- done. In particular, currently-commented out test script sections should be reviewed (deleted, uncommented, or adapted) along with more general assessment and improvement.

https://github.com/NOAA-GFDL/FRE-NCtools/issues/74

File: t/Test10-remap_land_res.sh [hunk 2]
```diff
@@ -79,103 +68,4 @@
 		--dst_cold_restart dst_cold_restart/land.res \
 		--remap_file remap_file_C48_to_C192 --print_memory
   [ "$status" -eq 0 ]
-
```
ngs333, line 82: @underwoo @ceblanton 
Are we getting rid of these other remap_land tests?

ngs333 (approved):

PR #77, opened 2021-04-09
Updated build instructions for GFDL sites using "eval `env.sh`" usage

### LINKED ISSUE #75: Update build instructions for case using site-config env.sh
The build instructions should be updated to reflect the recent site-config env.sh changes.
More specifically, they should be augmented (with commands) for the case of  building on 
some GFDL sites where the site-config env.sh can be used.

ceblanton:
Closes #75


ngs333:
I've noticed that some folks do not immediately notice that the quotes are single forward and not something else.  Do you think it might be a good idea to someplace sneak in a word/text hint about this?

ceblanton:
Thanks, great point! Both the backquotes and the eval could use a note of explanation. I added an explanation directly above the `eval` for simplicity (it could go someplace else)

README.md
```diff
@@ -81,12 +81,21 @@ make
 make install
 ```
 
-Additionally, you will probably need to set the environment variable CONFIG_SITE
-and set the recommended environment (modules):
+If you are building on some GFDL-managed systems, you can load recommended
+environment modules (and the associated recommended autoconf defaults)
+by running the output generated by the site-specific `env.sh` script,
+located in the [site-configs](site-configs) directory:
```
underwoo, line 87: Suggested change:

```
If building on a GFDL-managed system, the recommended environment with the recommended autoconf defaults can be loaded by running the output generated by the site-specific `env.sh` script, located in the [site-configs](site-configs) directory:
```

underwoo (changes requested):

README.md
ceblanton: Updated, thanks Seth

ngs333 (approved):

PR #81, opened 2021-04-29
Remap land newdata

ngs333:
The PR is for am update to the remap_land app to remap lm4p (lm4.1 and lm4.2) data. There are four two requirements of this software : I)) The processing of new fields and/or dimensions used for lm4p, and  II) thet processing of a true 4D cohort_index. There are also three “weaker” requirements III) the memory consumption of the application, particularly given the potentially huge size of the cohort index, IV) execution speed of the application,  and V) ease of maintenance of the application.

I) New fields and dimensions

The lm4.0 version supported the dimensions of time,  lat, lon, zfull, tile, tile_index, cohort,  and  cohort_index. There are numerous fields, and fields were allowed to depend on these combinations of dimensions:
```
[time]
[<time> lat] [<time> lon]
[<time>, zfull, tile_index]
[ <time>,tile_index]
[<time>, cohort_index] 
```
where the time “dimension” is optional and is in fact not a true netcdf dimension except for the time field itself. Additionally, the length of the cohort dimension is one and the ```tile_index ``` (a compressed three dimensional index of components [tile,lat,lon] ) is used to represent the 4D cohort_index of ``` [<time>, cohort, tile, lat, lon] ``` .
For lm4p, the length of the cohort dimension can be greater than one, and  additional dimensions of nspecies, textlen, soilCCohort, litterCCohort and mdf_day exists.
There are numerous new fields, and they can be represented with the original lm4 dimension combinations above plus these:
```
[<time>,soilCCohort, zfull, tile_index]
[<time>, literCCohort]
[<time>, nspecies, textlen]
[<time>, mdf_day, tile_index]
```

II and III) Compressed cohort index size and and design
The previous remap_land authors had left comments in code to make a minimum memory version of the software. Now that for lm4p the cohort dimension can be greater than one, this became more important, and this version of the code keeps  cohort index compressed at all times as well as the 4D data that depends on that index. The tile and cohort dimension are fairly sparse.  With a common cohort dimension length of 80, and a tile dimension length of 22,  with a 96 grid the uncompressed cohort index would be of size 80 * 22 * 96 * 96 ~= 16 million. With three integer and one double array, this results in approximately 300 Meg of space assuming one core is used. With the current higher resolution grids an uncompressed index is not usable on some hardware, and the problem will be more acute in the future.  For this case the 4D index and data is kept completely compressed. (No attempt was made to change the software creating and using the (smaller) 3D tile_index and the data that directly dependents on it). 
The 4D index is created (see function compute_dst_cohort_index) as a search problem in the space of possible index values. Constraints such as the 3D tile_index is used to identify the valid 4D indexes - a valid cohort_index value necessarily has has a valid tile_index value embedded as three of the four dimensions. 

In the new land data there are well over a hundred (non-index) fields on the vegn2 file. As in the previous version of the code, the actual field is not read only needed. One field is read at a time, then it is remapped then written. The same memory space of the field data is used to store the next field. 

IV) Execution time.
 The execution time of the program is dominated by function full_search_nearest, which performs npoints_dst * npoints_src * nfaces great circle distance functions. This is the functionality of the highest computational complexity in the remap_land application. Mapping a C96 to a C96 grid takes about two to four minutes on one CPU. Mapping a C768 to C1152 would take about 8*8*12*12=9216 times as long. This is currently helped by the concurrency of the remap_land_parallel app, but 
it seems that complexity of this function should be reduced. If the great circle distance is a true metric distance, then the complexity can be easily improved by a factor on ( npoints_src/log (npoints_src) ) with metric trees. Otherwise, more complex geometric search techniques should be considered.

V) Maintenance.
At the start of this project it was estimated that the fastest course of action was to follow the style and technique of the previous authors, followed by an (eventual) C++ version. This was probably a mistake as Modern C++ provides decreased need of memory use code, many data structures, and many other programming conveniences and hemp that though independently they are minor they add up to a lot. Overall, the code would be much smaller and easier to develop and maintain. If a change is made to use modern C++, this would preclude the use of old compilers.

--------------------------------------------
Testing perform and missing testing.
the remap_land app was tested with lm4p data provided by Serge y Malyshev.  A total of 54 files are remapped, including the vegn2 and vegn1 files that contain the new vegetation type data, including data that depends on the 4D cohort index.  Fire data was also remapped.  The results of the remapping were used as IC for the new land model, and its results were in turn analyzed  by Sergey. The parallel version of remap_land was tested with number of pes equal to 1, 2, 4 and 8. The results were checked by comparing with nccmp each of the 54 filers to the corresponding one produced by the serial remap_land app.
Missing testing I:
The code contains software to process data that contains an optional time dimension. This has not been tested.
Missing testing II:
Additional CI testing scripts were not created yet.


ngs333:
@sml7826 I'm trying to add you and Meiyun as reviewers but have not been able to. Can you add yourselves?

ngs333:
A VIRTUAL MEETING FOR  PUBLICREVIEW OF THIS CODE IS POSSIBLE IF ANYONE WANTS IT. LET ME KNOW.

GFDL-Eric:
> A VIRTUAL MEETING FOR PUBLICREVIEW OF THIS CODE IS POSSIBLE IF ANYONE WANTS IT. LET ME KNOW.

@ngs333 I think this is a good idea. Let's schedule something for anyone who is interested.

ngs333:
[run_remap_land.txt](https://github.com/NOAA-GFDL/FRE-NCtools/files/6428488/run_remap_land.txt)

tools/remap_land/remap_land.c
```diff
@@ -267,1226 +315,2000 @@ int main(int argc, char* argv[])
...
       mpp_end_def(fid_dst);
 
-      
       /*-------------------------------------------------------------------------------
-	Remap the data and write out to dst_restart_file
-	It is assumed all the fields need to be remapped will have the first dimension
-	name "tile_index" or "cohort_index" ( excludes "tile_index" or "cohort_index")
-	-----------------------------------------------------------------------------*/
+        Remap the data and write out to dst_restart_file
+        It is assumed all the fields need to be remapped will have the first
```
ngs333, line 1696: This comment is not correct. E.g. Field dims can have time dim as first dim.

tools/remap_land/remap_land.c
```diff
@@ -1549,228 +2415,571 @@ int main(int argc, char* argv[])
...
+    void gather_compressed_int_data ()
+   ------------------------------------------------------------------------*/
+  void gather_compressed_int_data(int npts, const int *data, int *data_global) {
+    mpp_gather_field_int_root(npts, data, data_global);
+  }
+
+/*
+  gather compressed int data and make sure its sorted. If its not sorted, sort and adjust the associated data.
+  TODO: See if data can be gatherd so that its in order of pe number, then such data does notneed to be sorted?
+*/
```
ngs333, line 2693: Improve comment.

ceblanton:
Some of the stderr runtime output's formatting has extra newlines:

```
Memuse(MB) at End of loop face_dst = 1
, min=109.836, max=109.836, avg=109.836
```
instead of

```
Memuse(MB) at End of loop face_dst = 0, min=93.3906, max=93.3906, avg=93.3906

tools/remap_land/remap_land.c
ngs333: Changed.

tools/remap_land/remap_land.c
ngs333: Done.

ngs333:
@ceblanton 
Concerning comment https://github.com/NOAA-GFDL/FRE-NCtools/pull/81#issuecomment-832915622 ,  the output line was changed as requested.

PR #82, opened 2021-05-07
Fix for logic to skip fregrid reference test for non-Intel compilers

ceblanton:
This bitwise test for fregrid is not expected to pass on
non-Intel compilers, so a test was added to skip the test
if the $CC environment variable was not set to "icc".
In GFDL environments, the CC variable is not set to "icc" when
the Intel compilers are module loaded, so this test is not ever running.
Instead the test should be skipped if "icc" isn't in the PATH.

### DIFF
--- a/t/Test24-reference_fregrid.sh
+++ b/t/Test24-reference_fregrid.sh
@@ -24,7 +24,7 @@
 
 @test "Test fregrid ocean data" {
 
-  if [ "$CC" != "icc" ]; then skip "Test fails reference check without icc"; fi
+  if ! type icc; then skip "Test fails reference check without icc"; fi
   if [ ! -d "Test24" ] 
   then
   		mkdir Test24


t/Test24-reference_fregrid.sh
```diff
@@ -24,7 +24,7 @@
 
 @test "Test fregrid ocean data" {
 
-  if [ "$CC" != "icc" ]; then skip "Test fails reference check without icc"; fi
+  if ! type icc; then skip "Test fails reference check without icc"; fi
```
underwoo, line 27: I don't think this is the solution we want.  It is possible to build w/o Intel, but have the intel module loaded (or have `icc` in the PATH.

The reason `$CC` is not set is automaker does not pass the build environment variables to the shell when running tests.  Where the change _should_ be made is to have the `Makefile.am` pass `CC` to the shell that runs the tests.  There should be something already in one of the `Makefile.am`s.

Even with that, this test will still not work as desired.  It is possible to set `CC` to the full path.  This would also fail the check.  We will also run into a time when even the Intel compiler may not bit-wise reproduce.  

I feel it is better to drop the bit-wise tests, or to add them as a special case (e.g. `make bitwise-check`).  This will leave this check up to the user -- as the user should know if bit-wise reproducibility is expected.  

Note, if we are going to continue bit-wise checking, we should get bit-wise checking for more compilers, and not just Intel.

underwoo (changes requested):
Based on my comment, I'm not in favor of continuing bit-wise test checking as currently implemented.   I'm really not keen on merging this PR as-is.

PR #85, opened 2021-05-07
Add env.sh for NESCC

### LINKED ISSUE #83: Add env.sh for hera
The use of lmod on hera does not allow a user to have two compiler modules loaded.  This causes a problem when building with the Intel compilers.  The Intel compilers use the version of `gcc` that appears first in `PATH` to determine what C standard to use when compiling.  FRE-NCtools is using the C11 standard.  The system GCC on hera is too old and doesn't have a full C11 implementation.  

To get past this problem, an `env.sh` similar to what is used on GFDL and gaea would help a user get past the issue.

underwoo:
This adds a working env.sh for use on NESCC systems.

Adds newer `gcc` to `PATH` for Intel to use C11 standards.

Fixes #83

### DIFF
--- a/site-configs/nescc/config.site
+++ b/site-configs/nescc/config.site
@@ -1,0 +1,32 @@
+#***********************************************************************
+#                   GNU Lesser General Public License
+#
+# This file is part of the GFDL FRE NetCDF tools package (FRE-NCTools).
+#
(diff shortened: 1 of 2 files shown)


ceblanton (approved):
Tested on hera, works well

ngs333 (approved):

PR #93, opened 2021-05-21
Update split_ncvar.pl shebang header to use first perl found in user's PATH

### LINKED ISSUE #92: split_ncvars.pl should use relative-path perl
**Is your feature request related to a problem? Please describe.**
split_ncvars.pl currently uses the system Perl (/usr/bin/perl) via its shebang. It would be better to use the first `perl` in the user's PATH, and would support more varied system environments.
```
#!/usr/bin/perl
```

**Describe the solution you'd like**
Change split_ncvars.pl's shebang to use the first `perl` in the user's environment:
```
#!/usr/bin/env perl
```

**Describe alternatives you've considered**

**Additional context**

ceblanton:
resolves #92

In particular, this will help PP/AN CentOS 7 transition

### DIFF
--- a/postprocessing/split_ncvars/split_ncvars.pl
+++ b/postprocessing/split_ncvars/split_ncvars.pl
@@ -1,4 +1,4 @@
-#!/usr/bin/perl
+#!/usr/bin/env perl
 #
 #***********************************************************************
 #                   GNU Lesser General Public License


ceblanton:
We will need to back-port this to the Bronx-18 tag and then patch the bronx-18 on-disk install locations, in order to have Bronx-18 work on the Centos 7 PAN nodes.

chanwilson (approved):

PR #107, opened 2021-06-24
Use input_file format for the output NetCDF file type if the format argument is not specified

ngs333:
Fregrid is modified to more closely work as indicated in its usage instructions. According to the
instructions: "When format is not specified, will use the format of input_file." Fregrid now temporarily stores the
netcdf format of the input_file file upon opening it, and if the case where the "format" argument is not specified, it uses
that stored format for the format of the output file. Function reset_in_format was added to mpp_io.

This PR is related to GFDL issue ticket #5019381 ("bronx-18: fregrid fails to remap tiled files with netcdf3/netcdf4 cconfusion") and was tested with the data indicated mentioned in that issue.

NOTE:: This PR replaces PR 106 because the automatic re-formatting created be Visual Code was too great.

### DIFF
--- a/tools/fregrid/fregrid.c
+++ b/tools/fregrid/fregrid.c
@@ -280,6 +280,8 @@ char *usage[] = {
 const double D2R = M_PI/180.;
 char tagname[] = "$Name: bronx-10_performance_z1l $";
 
+extern int in_format; //declared in mpp_io.c
+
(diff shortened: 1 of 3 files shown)


ceblanton (approved):

ceblanton:
Thanks @ngs333, I also tested on PP/AN. Merging in

PR #118, opened 2021-10-07
Supergrids from internal FV3 grids.

ngs333:
The PR modifies make_hgrid to correctly make supergrids from internal FV3 grids. Make_hgrid already contained a procedure to
read NetCDF files that contained  the four fields grid_lon, grid_lont, grid_lat, and grid_latt. These fileds would be read, the points (lat and lon coordiantes) would be used as supergrid cell verticies and centroids, and the supergrid cell face midpoints would be calculated from them.
To correctly make this work, these changes were made to make_hgrid  :
a) create_grid_from_file() function was modified so that the angular midpoints (lat and lon) between two points on the sphere (actually, two neighboring cell vertices) are the angles corresponding to the spatial midpoint on the great circle. The old method calculated the mid latitude as a simple 0.5 * (angular sum of neighbor latitudes); and similarly for the longitudes .
b)  make_hgrid() function was modified to correcly allocate space for the supergrid when more than one tile is specified  (in this case where the grid_type is "from_file").

Finally, a legacy flag was added as an "already deprecated" feature  (to be removed in a future PR) to temporarily allow the use of the old/legacy method for calculating midpoints. (the new method is the default method)

### DIFF
--- a/tools/make_hgrid/create_grid_from_file.c
+++ b/tools/make_hgrid/create_grid_from_file.c
@@ -50,7 +50,7 @@ void get_grid_v5(int fid, double *xt, double *yt, double *xc, double *yc, int ni
    cell area and rotation angle
 ************************************************************************/
 void create_grid_from_file( char *file, int *nlon, int *nlat, double *x, double *y, double *dx, double *dy,
-           		    double *area, double *angle_dx, int use_great_circle_algorithm )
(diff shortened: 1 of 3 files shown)


bensonr:
@ngs333 - short of finding some CM2.1 or ESM2 datasets and mosaics, that's probably the easiest way to do it.

PR #119, opened 2021-10-07
Small tweaks for the ocean model grid generator tests

ceblanton:
- call the tool as a command rather than through python3
  (as the user would call it)
- call the tool directly as opposed to through bats "run"
  which hides the standard out and does not exit on error
- remove unneeded exit code check which caused a spurious test failure


ngs333:
Issue shoudl be resolved in https://github.com/NOAA-GFDL/FRE-NCtools/pull/122

PR #122, opened 2021-10-09
Ocean grid model.subproject

underwoo:
Convert `tools/ocean_model_grid_generator` to a autoconf subproject.  Needs to have NOAA-GFDL/ocean_model_grid_generator#2 and adcroft/numpypi#2 merged eventually.


underwoo:
Need [upstream PR](https://github.com/nikizadehgfdl/ocean_model_grid_generator/pull/12) to be addressed before merging this PR.  Also, need to update the submodule URL and hash.  At this point, the submodule is pointing to my form of ocean_model_grid_generator.  This was done to ensure the tests will pass with the updated docker container.

underwoo:
@ngs333, my guess is you didn't get the submodule data.  You need to either clone with the `--recursive` option, or run `git submodule update --init` before running `autoreconf -i`.

ngs333:
@underwoo 
Yes looks like I missed the "submodule update --init" before "autoreconf -i".
But after that, running configure without any arguments yields:
```
checking python3 module: numpypi... no
configure: error: failed to find required module numpypi
configure: error: ../../../tools/ocean_model_grid_generator/configure failed for tools/ocean_model_grid_generator
```

ngs333:
@underwoo 
BTW, I can manually run this command 
```python3 -m pip install git+https://github.com/adcroft/numpypi.git```
In which case configure can be re-run and completes.

**Comment by underwoo** [comment]
@ngs333: yes, the configure script for `ocean_model_grid_generator` checks if all the required python modules are available for the python used.  What you saw is expected behavior.  If your python doesn't have the required modules, you can either `pip install` as you did, or pass the option `--enable-venv` to configure.  This will install a python virtual environment that will be used to run the ocean model grid generator script.  You can see this option in the FRE-NCtools configure if you run `configure --help=recursive`.

**Comment by ngs333** [comment]
@underwoo 
I tried using your branch as below, but I ran into some unexpected behavior.
First I got a version of your branch merged onto the baseline like this:
```
$git clone --recursive https://github.com/NOAA-GFDL/FRE-NCtools.git
$cd FRE-NCtools
$git remote add seth https://github.com/underwoo/FRE-NCtools.git
$git fetch seth
$git pull seth ocean_grid_model.subproject
$git submodule update --init --recursive
```
(Result was Warning: unable to rmdir 'numpypi': Directory not empty
Submodule path 'tools/ocean_model_grid_generator': checked out '7c7fb4ddefddc9ce81558f13dbe48cdd1235a7d3')
The last step above step was repeated after I removed the contents of the directory numpypi, with no warnings. I think 
this wont be an issue after the merge of the PR)
$autoreconf -i

I tried these four cases below. Comments are in parenthesis:

I)
```
$configure --disable-ocean-model-grid-generator
$make && make check
```
(this above works )

II)
``` $configure ```
( this above fails with error:
checking python3 module: numpypi... no
configure: error: failed to find required module numpypi
) 

III)
```
$configure --enable-venv  (retest this one from scratch)
$make
```
(make works)
```$make check```
(make check stops after this failure:
PASS: test_funcs.py
FAIL: test_ocean_grid_gen.py
Also,  it takes a long time to run test_ocean_grid_gen.py)

IV)
```$configure --enable-venv --enable-ocean-model-grid-generator
$make
$make check
````
( Case IV results the same as case III above)

I am trying cases  III) and IV) on different hardware fight now. Also, should case II) (no arguments to configure)
default to case I)

Thanks

**Review by ngs333 — APPROVED** [review]

**Comment by ngs333** [comment]
@underwood @ceblanton 
Here is a status of trying code of this PR,  with configure options --enable-venv
I) On my GFDL workstation, make && make check  works. 
(I checked the subproject python tests and the pass - I can see :
PASS: test_funcs.py
PASS: test_ocean_grid_gen.py
)
II) On analysis. I am stuck but I think there are network issues (see posts above: https://github.com/NOAA-GFDL/FRE-NCtools/pull/122#issuecomment-1069658792)

III) On gaea, on the 2nd run of "make check" , it runs the tests but   test_ocean_grid_gen.py reports a failure.
     The first time I ran "make check" it terminated early with this messages
```
  Running setup.py install for netCDF4 ... error
    ERROR: Command errored out with exit status 1
...
src/netCDF4/_netCDF4.c:980:10: fatal error: netcdf.h: No such file or directory
     #include "netcdf.h"
```

underwoo:
@ngs333 I don't expect the `--enable-venv` to work from the analysis nodes, as they do not have open access to the internet.  

Also, the ocean_grid_model tests will not always pass.  They currently use checksums that are not reproducible on all systems.  I have reported the issue to @nikizadehgfdl.  At this point, it is not clear if the numpypi python module (which is developed by @adcroft, and is essentially a wrapper over the numpy module with the goal to be reproducable), or how something else in the ocean_model_grid_generator source.  My suggestion at this point is to accept this PR.  Then if any other issues arise, we can track the specific problems in specific issues.

PR #104, opened 2021-06-23
Testing script and CI updates

rem1776:
Updates testing scripts with a new file for default setup and teardown functions and functions for generating input files.
Also adds a way to skip tests via an environment variable `SKIP_TESTS` and adds parallel tests to run when built with MPI. 

Fixes #74

PR #124, opened 2021-10-14
Two test updates that didn't make it in #104

ceblanton:
- Skip the Test04 parallel tests only if requested (i.e. potentially run through CI)
- Compare the serially and parallel fregrid output


ceblanton:
@ngs333, this adds a nccmp check for the parallel part of Test15 and uses different (improved?) logic for skipping the slow Test04 on the github runners (as it's way too slow). I wanted to allow the option of running the parallel part of Test04 through CI *and* skip it in the github runners. It works, but not sure it's a good approach. @rem1776 what do you think?

PR #129, opened 2021-11-24
Writing uniform provenance info in netdfs global attributes.

ngs333:
This PR creates print_provenance utility functions to uniformly (accross all NCTols apps using the functions) write as netcdf global attributes provenance and provenance related information.
A sample of the info is 
```
// global attributes:
		:grid_version = "0.2" ;
		:code_release_version = "19.0.1" ;
		:git_hash = "51ed4692c8b238e1e15bbc1d240951d3000e6f35" ;
		:creationtime = "Wed Nov 24 13:52:24 2021" ;
		:hostname = "edison" ;
		:history = "make_hgrid --grid_type gnomonic_ed --nlon 96 --grid_name C48_gridy" ;
```
The new functions are in /tools/libfrencutil/tool_util.c.
configure.ac was modified to place the git_hash info into the generated config.h file.
Numerous apps were modified to use the new functionality.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -26,6 +26,10 @@ m4_define([git_revision],
   [m4_esyscmd_s([git describe --always --dirty 2> /dev/null || echo "unknown version"])])
 AC_REVISION([$git_revision])
 
+m4_define([git_hashval],
+  [m4_esyscmd_s([git rev-parse HEAD 2> /dev/null || echo "unknown githash"])])
+AC_REVISION([$git_hashval])
+
 AC_CONFIG_SRCDIR([postprocessing/land_utils/combine-ncc.F90])
 AC_CONFIG_MACRO_DIR([m4])
 AC_REQUIRE_AUX_FILE([tap-driver.sh])
@@ -133,6 +137,8 @@ AM_CONDITIONAL([HAVE_TEST25_INPUT], [test -e ${srcdir}/t/Test25-input/grid_spec.
 # Other defs
 AC_DEFINE([GIT_REVISION], "git_revision",
           [Holds the 'git describe' information if configure ran within a git working directory])
+AC_DEFINE([GIT_HEADHASH], "git_hashval",
+          [Holds the 'git rev-parse HEAD' information if configure ran within a git working directory])
 # Output files
 AC_CONFIG_HEADERS([config.h])
 AC_CONFIG_FILES([Makefile
--- a/configure.ac
+++ b/configure.ac
@@ -20,7 +20,7 @@
 
 # Prelude
 AC_PREREQ([2.63])
-AC_INIT([FRE NC Tools], [18.0.0], [oar.gfdl.help@noaa.gov], [fre-nctools], [https://github.com/NOAA-GFDL/FRE-NCtools])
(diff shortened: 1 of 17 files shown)


ngs333:
@ceblanton 
Hi Chris, I havent set reviewers yet because I am still testing this. Quite a bit of generated artifacts are changed.

bensonr (approved):

ceblanton:
Might it be better to include the full hostname? e.g. `an201.princeton.rdhpcs.noaa.gov` instead of just `an201`? I was thinking of perhaps a case like `login1.school.edu` where the short hostname might be ambiguous.

In almost every case it would make no difference, so I'm fine with the short name as is.

ceblanton (approved):

nikizadehgfdl:
What code does code_release_version = "19.0.1"  refer to?   fre-nctools tags are like v2.*
If precompiled bronx-19 is used what would  appear for code_release_version? 
Who sets the grid_version?

ceblanton:
@nikizadehgfdl, the code_release_version comes from the `configure.ac` file so will need to be updated on each release. The more authoritative state of the code is the `git_hash` value, which is the hash of the most recent commit. The git tags such as `v2.18.0` should be the same as `git describe <git_hash>`.

No idea about the grid_version. I'd like to think it's like a mini-version unique and meaningful to each grid tool, and is updated by each tool as needed.

ngs333:
> What code does code_release_version = "19.0.1" refer to? fre-nctools tags are like v2.* If precompiled bronx-19 is used what would appear for code_release_version? Who sets the grid_version?

The code release version is to be set by the user compiling the code with autotools. 
In confiure.ac file, its the 2nd argument in AC_INIT.

I'm still working on the meaning of the grid-version, but there seems to be two types:
A) Most apps (make_hgrid, make_vgrid, make_topog, make_[quick,coupler,solo]_mosaic astually just set it to 0.2
and write it to the nc file.

B) check_mask app and the functions in  tool_util.c use a variable named grid_vesion which can equal to one of three things. What it could mean may be given by this code :
```
if(mpp_field_exist(grid_file, "wet") ) {
    grid_version = VERSION_1;
    printf("\n ===>NOTE from check_mask: the grid file is version 1 grid which contains field wet\n");
  }
  else if(mpp_field_exist(grid_file, "gridfiles") ) {
    grid_version = VERSION_2;
    printf("\n ===>NOTE from check_mask: the grid file is version 2 (solo mosaic grid) grid which contains field gridfiles\n");
  }
  else if(mpp_field_exist(grid_file, "lnd_mosaic") ) {
    grid_version = VERSION_3;
    printf("\n ===>NOTE from check_mask: the grid file is version 3 (coupled mosaic grid) grid which contains field lnd_mosaic\n");
  }
```
and some logic is modified depending on which version it is.
It seems to me type A) and B) are not related, but i'm not 100% sure.

nikizadehgfdl:
Thanks. My point was that where/which repo would I find the  "19.0.1" tag?  fre-nctools repo or fms repo?
grid_version is tricky. I kind of remember the VERSION_1 and VERSION_2 in B above to signify pre-mosaic and post-mosaic. But 0.2 does not ring any bell and is confusing.

PR #137, opened 2021-12-08
Draft GFDL gitlab updates for deployment

ceblanton:
Do not merge!

While we want to enable deployments using the gitlab CI, it doesn't feel safe with the settings as is. i.e.:

The version to be installed to is set in configure.ac, but the tag is the gitlab trigger, so with a new tag, the existing old tag installation location would be overwritten.

I think it would be safer to install the location to the git tag instead of the configure.ac package version if possible.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -13,7 +13,7 @@ variables:
     - autoreconf -i
     - mkdir ${buildDir}
     - cd ${buildDir}
-    - ../configure --prefix=${FMS_HOME}/local/opt/fre-nctools/test/${FRE_SITE}
+    - ../configure
     - make
     - make -j check
   artifacts:
@@ -73,6 +73,7 @@ deploy:gfdl-ws:
     FMS_HOME: /home/fms
   only:
     - master
+    - tags
   dependencies:
     - build:gfdl-ws
   <<: *deployScript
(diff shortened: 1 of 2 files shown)


ceblanton:
@ngs333, after verifying, these changes are not ideal and we should discuss a bit. I used the settings for deploying the Bronx-19 patch, so they work, but they require some special care or else may deploy the tag releases to the "test" location by accident.

The unobjectionable update is running the deploy script on gitlab CI on new tags.

The unpalatable situation is how they will be deployed at GFDL. Removing the --prefix option from the ./configure in the .gitlab-ci.yml allows autoconf to use the default in the <site>/config.site file: e.g. for GFDL/PAN,

`test "$prefix" = NONE && prefix=/home/fms/local/opt/fre-nctools/${PACKAGE_VERSION}/gfdl`

The wrinkle is that PACKAGE_VERSION is set in `configure.ac`, not the git tag as we want. This requires setting PACKAGE_VERSION to the tag, or else it will install to "test" by default.

Better would be to set PACKAGE_VERSION to the git tag somehow. @underwoo do you know how we could do that, and is that a good idea? Any other thoughts on this? Thanks!

ceblanton:
Closing this. I have a better idea

Issue #47, opened 2021-02-12
run.cr_simple_hydrog.csh fails and wipes the entire FTMPDIR


ngs333:
When I try to run run.cr_simple_hydrog.csh from a sub-directory of $FTMPDIR, it fails with the following messages:
an106: /vftmp/Sergey.Malyshev/pid20895/hydro > ./run.cr_simple_hydrog.csh 
ERROR: invalid platform
In addition, everything in the $FTMPDIR directory is wiped out by the script. This could potentially result in significant loss of work, if there are other directories in $FTMPDIR that are being used for, say, data accumulation or processing. The "invalid platform" error message appears to be spurious, caused by the fact that in my case the current working directory was also removed.
Ideally, a script should create its own unique temporary directory, if necessary.
@slm7826
@ngs333 This is gitlab issue #65

ceblanton:
Hi @ngs333, @LuisSalbey and I modified the original `run.cr_simple_hydrog.csh`, and renamed it `make_simple_hydrog.csh` in the process. The script included in this repo is the `make_simple_hydrog.csh` currently.

ceblanton:
However, we didn't fix that bad behavior you see, unfortunately. I agree it is suboptimal and even dangerous.

ngs333:
@slm7826
Hi Sergey. Sorry we took so long to get to this issue. The original script ( run.cr_simple_hydrog.sh) was updated a while ago, and now renamed make_simple_hydrog.sh. The new script is in dir FRE-NCtools/tools/simple_hydrog;  and it seems the old script was somehow deprecated and yet kept and and put in dir FRE-NCtools/tools/simple_hydrog/old.

I agree its bad to wipe out the directory. It looks like both scripts do that, and the new one will be changed this week. Possibly to just make a new dir with mktemp and to work from there.

I am wondering if you are still using, want to use, or know someone that is using the old script.
Thanks
Mike Z

slm7826:
Hi Miguel,

I have not used this script since the time I submitted the ticket; I am not sure if anyone is using the old version. 

Not that it would help you very much, but I think the most likely people to use this script (old or new) are the ones who create new grid specs and want to run something quick without going through manual creation of the hydrography. Not sure how to identify them though.

--
Sergey

ngs333:
@slm7826 
Sergey,
Thanks for the input.
MZ

PR #148, opened 2022-03-04
make_simple_hydrog.csh and unit test data links.

ngs333:
This PR
1) changes script make_simple_hydrog.csh so that it does not delete temp directories and instead make
    a new temp direcotry as suggested by S. Malyshev (see https://github.com/NOAA-GFDL/FRE-NCtools/issues/47)
2) Makes fixes to two links to directories accesible on analysis in /home/fms/fre-nctools so that at least 
    the simple_hydrology unit test (currrently Test25) can be run.

Testing:
With this change:
 - Test25 passes on Analysis host. 
 - Test25 is skipped in CI and it is skipped in home computer installation.
 -


ngs333:
@ceblanton @underwoo @rem1776 
Wiith the proposed changes, Test25 passes on Analysis but not on CI. Given that the script (and earier versions in other dirs and sites) has always indicated that it should be run on Analysis, I am tempted to make the configuration (SKIP_TEST line in file FRE-NCtools/.github/workflows/mail.yaml) indicate to skip test 25. Is this the right approach?

ceblanton:
If we skip it only in the CI it will fail on other sites such as hera won't it?

How about if the test always runs, but checks if `gfdl_platform` is in the PATH. If not, it passes (exit 0) right away; if so, it runs the test.

ngs333:
I added the check for command gfdl_platform, which makes sense to me.  However, the CI test is
failing with this error:

```make[2]: *** No rule to make target  'share/river_network_mrg_0.5deg_ad3nov_fill_coast_auto1_0.125.nc', needed by 'all-am'.  Stop.```

Looking at file FRE-NCtools/tools/simple_hydrog/Makefile.am , that nc file above is a data target to be distributed. So as the CI system can’t find the target, it fails the build.I don;t know if thre is a way to make the data targets optional? Perhaps an altogether different approach is better.

**Comment by rem1776** [comment]
@ngs333 Just put the pr in for the fix, ended up being simpler than i thought because I must've added a configure check already:

https://github.com/NOAA-GFDL/FRE-NCtools/blob/f5d700467f100b63bf5bf92d69c76be989dd2924/configure.ac#L129-L130

So just adding the `if` seems to have fixed the CI failure

File: tools/simple_hydrog/share/make_simple_hydrog.csh [hunk 1]
```diff
@@ -31,19 +31,11 @@
...
-    if ($?TMPDIR) then
-        echo WARNING
-        echo "WARNING: Intended for PP/AN use; may not work properly on this site"
-        echo WARNING
-        echo
-    else
-        echo 'FATAL: $TMPDIR needed'
-        exit 1
-    endif
+    echo "WARNING: Intended for PP/AN use; may not work properly on this site"
```
ceblanton, line 34: I think we want the warning to print if the string does not match. But we also want to check for the `gfdl_platform` not being in the path. So think this will work:

```
if ( ! { gfdl_platform} || `gfdl_platform` != "hpcs-csc") then
    echo "WARNING: Intended for PP/AN use; may not work properly on this site"
endif

File: tools/simple_hydrog/share/make_simple_hydrog.csh [hunk 2]
```diff
@@ -31,19 +31,11 @@
...
 endif
 
+set my_tempdir=`mktemp -d`
```
ceblanton, line 37: Suggest checking the exit status in case user can't write to TMPDIR:

```
if ( $status ) then
    echo ERROR: Cannot create temporary directory
    exit 1
else

**Review by ceblanton — CHANGES_REQUESTED** [review]
I left two small cshell related comment suggestions above.

There are two needed data files, right? You re-added the link for river_network_mrg_0.5deg_ad3nov_fill_coast_auto1_0.125.nc , does the other one need to be added too?

```
gigbp2_0ll.nc -> /home/fms/fre-nctools/share-datasets/gigbp2_0ll_v001.nc

ngs333:
> I left two small cshell related comment suggestions above.
> 
> There are two needed data files, right? You re-added the link for river_network_mrg_0.5deg_ad3nov_fill_coast_auto1_0.125.nc , does the other one need to be added too?
> 
> ```
> gigbp2_0ll.nc -> /home/fms/fre-nctools/share-datasets/gigbp2_0ll_v001.nc
> ``

@ceblanton
For running make_simple_hydrog.sh the link to river_network_mrg_0.5deg_ad3nov_fill_coast_auto1_0.125.nc
is not needed. Neither will be the link to  gigbp2_0ll_v001.nc. The reason the script has access to the shared library dir anyway, and the files are found relative to (inside) that dir. 
Also, I don't really know what the makefile.am has it as a target to export. But if there are circumstances which that export is needed, then adding a link for  gigbp2_0ll_v001.nc would also be a good idea.

tools/simple_hydrog/share/make_simple_hydrog.csh
ngs333: OK. I'll add the ```! { gfdl_platform} || ``` though I do not think it makes a difference.

tools/simple_hydrog/share/make_simple_hydrog.csh
ngs333: OK. Will add.

ceblanton (approved):
Looks good. If we need to add symlinks later we can

PR #151, opened 2022-04-01
README.md for  2022.1 - 2nd attempt.

ngs333:
This is an update to the README.md file. It includes additions related to the recently merged changes do to the ocean_model_grid_generator subproject, plus miscellaneous improvements.

### DIFF
--- a/README.md
+++ b/README.md
@@ -61,9 +61,9 @@ and also it is a Python3 project. Because of the former attribute, recursive
 cloning (see below) is recommended. Because of the latter attribute,
 the users python environment may need modification and/or options to the
 autotools configure command may need to be specified. This includes the
---[enable|disable]-ocean-model-grid-generator option and the --enable-venv
-option to build a Python venv containing the Ocean Model Grid Generator script
-and all python dependencies.
+--[enable|disable]-ocean-model-grid-generator option (default is enable)
+and the --enable-venv option to build a Python venv containing the
+Ocean Model Grid Generator script and all python dependencies.
 
 ## Cloning and submodules
 The NCTools github repository contains Ocean MOdel Grid Generator's repository
--- a/README.md
+++ b/README.md
@@ -53,8 +53,20 @@ The tools available in FRE-NCtools are:
 * runoff_regrid -- Regrid land runoff data to the nearest ocean point
 * transfer_to_mosaic_grid -- Convert older style grids to newer mosaic grid
 
+### Other tools
+The [ Ocean Model Grid Generator](https://github.com/NOAA-GFDL/ocean_model_grid_generator)
+is a collection of tools for creating finite element spherical tripolar grids for
+GFDL's MOM based ocean models. Unlike the other tools, NCTools includes it as a submodule,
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -53,8 +53,20 @@ The tools available in FRE-NCtools are:
...
+cloning (see below) is recommended. Because of the latter attribute,
+the users python environment may need modification and/or options to the
+autotools configure command may need to be specified. This includes the
+--[enable|disable]-ocean-model-grid-generator option (default is enable)
+and the --enable-venv option to build a Python venv containing the
+Ocean Model Grid Generator script and all python dependencies.
+
 ## Cloning and submodules
-The NCTools github repository contains the GFDL ocean_model_grid_generator repository
+The NCTools github repository contains Ocean MOdel Grid Generator's repository
```
underwoo, line 69: lowercase the `O` in `Model`.

README.md
```diff
@@ -53,8 +53,20 @@ The tools available in FRE-NCtools are:
 * runoff_regrid -- Regrid land runoff data to the nearest ocean point
 * transfer_to_mosaic_grid -- Convert older style grids to newer mosaic grid
 
+### Other tools
+The [ Ocean Model Grid Generator](https://github.com/NOAA-GFDL/ocean_model_grid_generator)
```
underwoo, line 57: Please remove the space after the `[`.  I usually wouldn't be this specific, but as there are other comments, might as well address this one as well.

README.md
```diff
@@ -53,8 +53,20 @@ The tools available in FRE-NCtools are:
...
+is a collection of tools for creating finite element spherical tripolar grids for
+GFDL's MOM based ocean models. Unlike the other tools, NCTools includes it as a submodule,
+and also it is a Python3 project. Because of the former attribute, recursive
+cloning (see below) is recommended. Because of the latter attribute,
+the users python environment may need modification and/or options to the
+autotools configure command may need to be specified. This includes the
+--[enable|disable]-ocean-model-grid-generator option (default is enable)
```
underwoo, line 64: Please place these options within backticks.  This will help identify them as options.  For example:

<pre>
This includes the `--disable-ocean-model-grid-generator` option ... .
</pre>

Also, when an option is needed that enables or disables, or with or without, then include the option that is not the default.  In this case, the option is `--disable-ocean-model-grid-generator`.

underwoo (changes requested):

README.md
ngs333: All improvements are appreciated!

README.md
ngs333: @underwoo 
OK.  Leat me know if the latest push needs further improvements. Thanks.

ceblanton (approved):
Personally I would find the --enable-venv option confusing as a user. Shouldn't --enable-omgg and --enable-venv be both turned on together or both both turned off together? (Or if --enable-omgg is on by default, shouldn't --enable-venv also be on by default?)

I don't want to hold this up for that, though. Users can apply the two options as documented as they see fit.

Issue #99, opened 2021-06-17
Switch GFDL site configurations to GCC


ngs333:
NCTools needs to document and set GCC compiler flags and settings for use after the 18.1 release.
A good set of flags will consider the parameters of reproducible with Bronx-18, speed of execution, and help in debugging working executables, and help in adding and maintaining code.

Both the config.site files and documentation may need changes.

ceblanton:
Currently, GCC 9.2.0 works OK on workstations and 10.3.0 works OK on gaea with minimal site changes. But there is no suitable GCC on PAN currently.

https://github.com/ceblanton/FRE-NCtools/tree/99.gcc

PR #153, opened 2022-04-21
Update GFDL compiler settings from Intel to GCC

ceblanton:
#99


ceblanton:
@ngs333 I tested the 3 configs and they all compiled and passed tests (except for the omgg). However, I have zero build options which is probably not right, and I don't know what we want.

site-configs/ncrc/config.site
```diff
@@ -29,10 +29,10 @@ test -z "$CC" && CC=cc
 test -z "$FC" && FC=ftn
 
 # Compile/Link flags
-test -z "$CFLAGS" && CFLAGS="-xSSE2"
```
ngs333, line 32: Chris, The Intel default optimization flag is -O2. The GNU default optimization flag is -O0.
So we need to add the -O2 flag whenever we use gcc. Thats for both CFLAGS and FCFLAGS.

ngs333 (changes requested):
The Intel default optimization flag is -O2. The GNU default optimization flag is -O0.
So we need to add the -O2 flag whenever we use gcc. Thats for both CFLAGS and FCFLAGS.

ngs333 (approved):

ngs333 (approved):

PR #159, opened 2022-05-06
Suggest target latitudes so that stretched grid would have both NS poles

nikizadehgfdl:
In general for a given stretch factor and target latitude the resulting stretch grid will not have the poles as grid points.
This may cause issues later with other tools such as exchange grid generator manifested as tiling errors and cells with a wrong land mask.
- Work related to issue https://github.com/NOAA-GFDL/FMS/issues/621 in FMS repo (wrong repo for that issue)
- Implement the condition for the stretched grid to have both poles as grid points for a give stretch factor.
- This subroutine lists possible target latitudes for a given stretch factor close to the requested target latitude such that the final grid has both poles as grid points.
- Having a grid that includes both poles potentially avoids later issues with exchange grids. After using the suggested values from this update no such issues were found.

### DIFF
--- a/tools/make_hgrid/create_gnomonic_cubic_grid.c
+++ b/tools/make_hgrid/create_gnomonic_cubic_grid.c
@@ -84,8 +84,8 @@ void cell_north(int ni, int nj, const double *lonc, const double *latc, double *
 void calc_cell_area(int nx, int ny, const double *x, const double *y, double *area);
 void direct_transform(double stretch_factor, int i1, int i2, int j1, int j2, double lon_p, double lat_p,
 		      int n, double *lon, double *lat);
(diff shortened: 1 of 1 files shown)


ngs333 (commented):
@nikizadehgfdl 
Is the functionality in suggest_target_lats() being used by someone now or expected to
be used as is ? If so, should it be an option (and also user documentation  in make_hgrid?
If not, should it just be commented out or possibly only use if compile-time flag is set?

nikizadehgfdl:
The functionality is not being used right now. There is potential benefits for the users to see the message. 
We could add an option to use a suggested value, but this is not the intent of this PR as there are multiple "good" values for target the latitude above and below the requested one. The intent is to provide the information in the output.

ngs333 (approved):

PR #164, opened 2022-05-18
Rpath for netcdf libs

ngs333:
This PR enables the addition of the netcdf libraries to the rpath of all NCTools compiled executables.
File configure.in and various Makefile.ac are modified, the first one to define and set an autotools variable for rpath, and the latter ones to use it.

This PR has been tested on Analysis by compiling NCTools (the canonical way using the site env.sh) ,
doing a module purge, and then making sure each executable  can start.


ceblanton (approved):
The update works; the executables now can run with no environment modification

ngs333:
@underwoo 
The failure was from new code that came in https://github.com/NOAA-GFDL/FRE-NCtools/pull/159.
That PR was reverted.

PR #165, opened 2022-05-20
Turn off ocean-model-grid-generator subtool for GFDL PPAN site

ceblanton:
PP/AN does not have open Internet access needed for virtual envcreating and testing.

Related to GFDL site file adjustments for 2022.01 (#99 )

### DIFF
--- a/site-configs/gfdl/config.site
+++ b/site-configs/gfdl/config.site
@@ -21,7 +21,8 @@
 #
 # configure options
 test -z "$with_mpi" && with_mpi=yes
-test -z "$enable_venv" && enable_venv=yes
+test -z "$enable_ocean_model_grid_generator" && enable_ocean_model_grid_generator=no
+test -z "$enable_venv" && enable_venv=no
 # Standard prefix location for FRE distribution
 test "$prefix" = NONE && prefix=/home/fms/local/opt/fre-nctools/${PACKAGE_VERSION}/gfdl
 


underwoo:
I have one question.  Why?  #99 does not explain why OMGG needs to be off for PPAN.

underwoo (commented):
Please let me know why OMGG needs to be off for PPAN nodes.

ngs333:
> Please let me know why OMGG needs to be off for PPAN nodes.

@underwoo 
@ceblanton attributed errors in setting up the venv to firewall settings for PPAN. He may be able to give
additional info when he returns.

underwoo:
@ngs333, thank you for the information.  Depending on what is blocked, we should be able to put in a CM request to open firewall ports.  I would rather go down that path than remove something the users will expect to be available on that platform.

ceblanton:
@underwoo, @ngs333 is correct. The PP/AN firewall ports are blocking some part of the python environment testing and creation. I agree opening the ports would be better.

chanwilson:
I don't see the details on what python module is attempting to be installed?

ceblanton:
These pip calls fail from the ocean-model-grid-generator package, when testing or installing the python virtual environment:

```
python3 -m pip install --upgrade pip
python3 -m pip install pytest
```

Yes, I believe access to the PyPI index (https://pypi.org/) will resolve this.

ceblanton:
Running the pip commands with verbose logging shows that they are trying to access two locations:

```
https://pypi.org
https://files.pythonhosted.org
```

@chanwilson and @underwoo, I believe I should request those two locations for port 443, then. Should I request port 80 as well?

underwoo:
@ceblanton, @chanwilson will know better, but to be safe please request both ports 80 and 443.  A quick google search didn't return anything specific about PyPi for firewall rules.

ceblanton:
Help desk ticket submitted. Let's close this, then.

PR #167, opened 2022-05-20
Revert "Suggest target latitudes so that stretched grid would have both NS poles"

ngs333:
Reverts NOAA-GFDL/FRE-NCtools#159
make_hgrid core dumps for certain inputs. See Test04-grid_coupled_nest.sh, second cal to make_hgrid, lines 70ff for an example set of args that can cause a core dump. In function suggest_target_lats, index ls in into array lat[] (line 1361 : 
``` b = -2*(1.0+sin(lat[ln])*sin(lat[ls]))/(sin(lat[ln])+sin(lat[ls])); ```)  may be negative and cause a core dump.


nikizadehgfdl:
The error suggests that my formula doesn't work with nested grids. It was not intended to work with anything but global stretched grids. I appears that I have to introduce an argument to go through the new subroutine only for global stretched grids.  Sorry about that.

PR #169, opened 2022-05-31
Suggest target latitudes so that stretched grid would have both NS poles

nikizadehgfdl:
- Implement the condition for the stretched grid to have both poles as grid
  points for a give stretch factor.
- This subroutine lists possible target latitudes for a given stretch factor
  close to the requested target latitude such that the final grid has both
  poles as grid points.
- Having a grid that includes both poles potentially avoids later issues with
  exchange grids.
- CI tests pass

### DIFF
--- a/tools/make_hgrid/create_gnomonic_cubic_grid.c
+++ b/tools/make_hgrid/create_gnomonic_cubic_grid.c
@@ -54,7 +54,8 @@
 #include "mosaic_util.h"
 #include "tool_util.h"
 #include "create_hgrid.h"
-
+#define  D2R (M_PI/180.)
+#define  R2D (180./M_PI)
 #define  EPSLN10 (1.e-10)
 #define  EPSLN4 (1.e-4)
 #define  EPSLN5 (1.e-5)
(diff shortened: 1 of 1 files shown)


tools/make_hgrid/create_gnomonic_cubic_grid.c
```diff
@@ -54,7 +54,8 @@
 #include "mosaic_util.h"
 #include "tool_util.h"
 #include "create_hgrid.h"
-
+#define  D2R (M_PI/180.)
```
ngs333, line 57: @nikizadehgfdl 
Instead of defining these here, can you #include "constant.h". Otherwise another warning is 
generated with certain compiler flags.

ngs333 (changes requested):

tools/make_hgrid/create_gnomonic_cubic_grid.c
ngs333: @nikizadehgfdl 
More specifically,  can you not define D2R and R2D and just ```#include "constant.h"```

nikizadehgfdl: @ngs333 thanks. I dropped the two defs and it worked. Pushed.

PR #192, opened 2022-11-17
Fix to runoff_regrid.c

ilaflott:
Test07-runoff_regrid.sh runs to completion on MacOS 12.5/Intel upon change to runoff_regrid.c.

### DIFF
--- a/.vscode/c_cpp_properties.json
+++ b/.vscode/c_cpp_properties.json
@@ -1,0 +1,22 @@
+{
+    "configurations": [
+        {
+            "name": "Mac",
+            "includePath": [
+                "${workspaceFolder}/**",
+                "/opt/homebrew/Cellar/netcdf/4.9.0/include"
+            ],
+            "defines": [
+                
+            ],
+            "macFrameworkPath": [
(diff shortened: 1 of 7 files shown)


ngs333 (approved):

.vscode/c_cpp_properties.json
```diff
@@ -0,0 +1,22 @@
...
+                "/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX.sdk/System/Library/Frameworks"
+            ],
+            "compilerPath": "/usr/bin/clang",
+            "cStandard": "c17",
+            "cppStandard": "c++17",
+            "intelliSenseMode": "macos-clang-arm64"
+        }
+    ],
+    "version": 4
+}
```
ceblanton, line 22: Was including this vscode config file intentional? It looks useful but might be specific to @ilaflott's setup.

underwoo: @ceblanton it was intentional.  It is specific to a mac using homebrew.  The configuration name should be more specific.  However, don't worry about it ATT.

ceblanton (approved):

PR #204, opened 2023-02-08
add initial openacc work

rem1776:
Adds the code changes from the gpu hackathon. includes some configure.ac changes to add the openacc flags for nvc after a simple check of the `-acc` flag (unless disabled via `--with-openacc=no`).

I tried to keep any changes that happened in between the hackathon and now in the code, but wasn't sure for certain things (mainly if the reproduce_siena flag and routines should be kept) so let me know if any removals look unrelated.

Flags were set up for the gpu box in the original code, I kept them the same except for `-ta=tesla`. Couldn't find any info on it in the nvidia docs so I changed it to `-gpu=cc70` which sets the 'compute capability' target to what the tesla gpu uses.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -17,6 +17,8 @@ jobs:
         MPI: ${{ matrix.with_mpi }}
         QUAD_P: ${{ matrix.enable_quad_precision }}
     steps:
+    - name: install autoconf dependency
+      run: apt update && apt install autoconf-archive
     - name: Checkout
       uses: actions/checkout@v2
       with:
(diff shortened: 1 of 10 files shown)


rem1776:
gonna open a new pr with less unneeded changes now that i have it compiling

ngs333 (commented):
This may not work as we have to be careful bringing in too many changes from  from the GPU work.
 Basically we want to start with the current FRE-NCTools/master baseline,  add a second version (from the GPU work) of the function create_xgrid_2dx2d_order2 in file create_xgrid.c, and only add or change code that is really needed to have that second function supported. And of course  we want to chose (preferably at compile time) between the two functions depending on compiler options . So I think probably only one, two , or three source should be affected: create_grid.c; and (possibly) create_xgrid.h;  and (possibly) mosaic_util.h. The latter only if we can't get nvc compilation (if we cant get around the 
Generic C lang feature in the fabs definition) through the choice of compiler options.

PR #205, opened 2023-02-13
Add in initial openacc code

rem1776:
See #204, replaces the code changes made there.
 
This compiles a separate version of libfrencutils and fregrid with the openacc flags `-acc -O2 -g -tp haswell -gpu=cc70 -Minfo=accel -Mnoinline`and names the executable `fregrid_gpu`.

I tried not bringing in too many unrelated changes from the existing code, but they were mostly unused variables/routines and trailing whitespace removals so it didn't really make sense to not keep those. Not really sure about the conserve_interp.c file though, wasn't going to include it but ended up just adding a debug macro in case we want it for benchmarking/testing.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -74,13 +74,13 @@ AS_IF([test $gx_cv_c__Generic = no],
 # TODO this should really be added as an m4 macro, surprised there wasn't a existing one in the archives
 AC_ARG_WITH([openacc],
   [AS_HELP_STRING([--with-openacc],
-    [Builds with openacc flags if available (default no)])])
+    [Builds with openacc flags for nvidias nvc compiler if available. This will result in a second executable for fregrid, fregrid_gpu. Flags added: -acc -O2 -g -tp native -gpu=ccnative -Minfo=accel -Mnoinline (default no)])])
 AS_IF([test ${with_openacc:-no} = yes],
   [with_openacc=yes],
   [with_openacc=no])
 AS_IF([test ${with_openacc} = yes],
   [AX_CHECK_COMPILE_FLAG( [-acc],
-       [ AC_SUBST(OPENACC_CFLAGS, "-acc -O2 -g -tp haswell -gpu=cc70 -Minfo=accel -Mnoinline")],
+       [ AC_SUBST(OPENACC_CFLAGS, "-acc -O2 -g -tp native -gpu=ccnative -Minfo=accel -Mnoinline")],
         AC_MSG_ERROR([OpenACC option specified but -acc flag not accepted by compiler]) )])
 AM_CONDITIONAL([WITH_OPENACC], [test "$with_openacc" = "yes"])
 
--- a/configure.ac
+++ b/configure.ac
@@ -70,6 +70,17 @@ GX_C__GENERIC
 AS_IF([test $gx_cv_c__Generic = no],
   AC_MSG_ERROR([The C compiler does not support the generic selection C-11 standard.  Please use a C-11 compliant compiler.])
 )
+# openacc support, simple check for working (nvc)openacc flag and allows for disabling if desired
+# TODO this should really be added as an m4 macro, surprised there wasn't a existing one in the archives
+AC_ARG_WITH([openacc],
+  [AS_HELP_STRING([--with-openacc],
+    [Builds with openacc flags if available (default yes)])])
+AS_IF([test ${with_openacc:-yes} = yes],
+  [with_openacc=yes],
+  [with_openacc=no])
+AS_IF([test ${with_openacc} = yes],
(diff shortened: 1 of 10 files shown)


tools/libfrencutils/create_xgrid.c
```diff
@@ -29,7 +29,8 @@
 #define EPSLN8            (1.e-8)
 #define EPSLN30           (1.0e-30)
 #define EPSLN10           (1.0e-10)
-
+#define R2D (180/M_PI)
```
ngs333, line 32: Constants like R2D and PTI were removed from here; they are in tools/libfrencutils/constant.h. Another reason to start with a copy of the baseline file.

rem1776: I did start with the baseline file and went through all these changes myself. Since the changes are from another repository there's no easy way of telling what was done in between now and the hackathon, especially since the entire file is formatted differently.

ngs333 (approved):

tools/libfrencutils/create_xgrid.c
```diff
@@ -887,6 +885,174 @@ int create_xgrid_2dx2d_order2_(const int *nlon_in, const int *nlat_in, const int
 
 };
 #endif
+#ifdef _OPENACC
```
mlee03, line 888: do we need an _openacc  macro since the directives will be ignored without the -acc compiler flag?

configure.ac
```diff
@@ -70,6 +70,19 @@ GX_C__GENERIC
...
+# TODO this should really be added as an m4 macro, surprised there wasn't a existing one in the archives
+AC_ARG_WITH([openacc],
+  [AS_HELP_STRING([--with-openacc],
+    [Builds with openacc flags if available (default no)])])
+AS_IF([test ${with_openacc:-no} = yes],
+  [with_openacc=yes],
+  [with_openacc=no])
+AS_IF([test ${with_openacc} = yes],
+  [AX_CHECK_COMPILE_FLAG( [-acc],
+       [CFLAGS="-acc -tp haswell -gpu=cc70 -Minfo -Mnoinline"],
```
mlee03, line 83: I think -Minfo=accel will give useful info about openacc compilation

tools/libfrencutils/create_xgrid.c
rem1776: Yeah it doesn't check for compiling its more for whats in the `#else`. It uses it to figure out if it should use the openmp or openacc version of the `create_xgrid_2dx2d_order2` function.

With this logic it'll also default to compiling the old openmp version if neither openmp or acc is enabled which I think is what we want.

configure.ac
ngs333: @mlee03 @rem1776 
The flags I was using for compiling for the gpu are:
 O2 -g -traceback -acc -ta=nvidia:managed -Minfo -Mbounds  -Mnoinline  

I think we may need to compare with the section "-tp haswell -gpu=cc70"

rem1776: The -tp is for the cpu architecture on the gpu box, not sure if thats technically needed.

-ta has been deprecated, -gpu is the new version with cc70 being the 'compute capability' that's on the tesla. After looking at the options a bit more there is also a `managed` option that can be added to the gpu flag, it turns on using CUDA managed memory. I'm not sure what that means entirely since we're not using cuda mallocs or anything like that in this code, but i'll test it out.

I'll definitely add in these:  `-Minfo=accel -O2 -g`

mlee03: I don't think we'll use the `managed` option after we optimize all the data transfers

configure.ac
```diff
@@ -70,6 +70,19 @@ GX_C__GENERIC
 AS_IF([test $gx_cv_c__Generic = no],
   AC_MSG_ERROR([The C compiler does not support the generic selection C-11 standard.  Please use a C-11 compliant compiler.])
 )
```
ngs333, line 72: @rem1776  @mlee03 
I can compile on the gpubox with nvc and without the configure option "--with-openacc". But if that option is added , then there are compilation warnings and  failures concerning the _Generic fabs declared in mosaic_util.h. I tried using the -std=c99, -std=c11, and -std=c17 as I have tried with other compilers, but in this case I get compilation messages like this:
```
"../../../tools/libfrencutils/create_xgrid.c", line 2296: warning: function "_Generic" declared 
     implicitly [implicit_func_decl]
```
Did anyone get past this problem in an acceptable way ?

mlee03: I get compilation errors too, it seems like the user defined CFLAG (and hence the -std=c11) gets overwritten

rem1776: I was wondering why I wasn't seeing this but looks like you figured it out, I was configuring using CC="nvc -std=c11" and it seemed to run fine cause it wasn't getting rid of the flag. That should be fixed now cause it'll add the openacc flags where needed instead of just setting them in the configure script.

PR #223, opened 2023-05-03
Remove the ocean_model_grid_generator subproject.

ngs333:
This PR has all the changes necessary for removing the ocean_model_grid_generator as a subproject to FRE-NCTools.


ceblanton (changes requested):
Looks good but there are intended `affinity.c` changes here also.

ceblanton (approved):

Issue #224, opened 2023-05-05
Add conda build recipe


ceblanton:
**Is your feature request related to a problem? Please describe.**
It would be nice to be able to install fre-nctools through miniconda as a secondary build method, in addition to building from source with autotools.

**Describe the solution you'd like**
A conda build recipe added, so one could build it with:
`conda build .`
Then, after either uploading to an anaconda channel or use a directory as a channel, install with:
`conda install fre-nctools`

**Describe alternatives you've considered**

**Additional context**

PR #225, opened 2023-05-11
Added conda build recipe and build script

ceblanton:
"conda build ." worked on the workstations, and conda install worked on the thin client.

#224

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -1,0 +1,4 @@
+autoreconf -i
+./configure --prefix=$PREFIX --disable-ocean-model-grid-generator
+make -j
+make install
--- a/meta.yaml
+++ b/meta.yaml
@@ -1,0 +1,22 @@
+package:
+  name: fre-nctools
+  version: 2022.02.01
+
+source:
+  path: .
+
+build:
+  number: 0
+
+requirements:
+  build:
+    - libnetcdf
+    - netcdf-fortran
+  run:
+
+test:
+  commands:
+
+about:
+  license: LGPL-3.0
+  summary: Tools for manipulating and creating netCDF inputs for FMS managed models


ceblanton:
Ideally, the conda build could happen through the github CI, and if successful, then it would publish to the noaa-gfdl anaconda channel:

https://anaconda.org/noaa-gfdl

For new releases (new git tags), we would want the published conda package to have the same version/tag.

For pushes to the main branch, we could have the published conda package have a `test` or `develop` version.

underwoo:
I think it is better to keep the conda recipe in a different repository. We
do not want to create a new conda binary package frequently, and only after
we mint a release.

ceblanton:
@underwoo thanks for the suggestion, but how is the frequency of conda binary releases related? Regardless of where the conda recipe is stored, we'll only release a conda binary once per release. I sort of see the advantage of keeping the recipes separate, in that build-only changes won't clutter up the source repository.

Hopefully, soon we'll have several conda and spack recipes, e.g. spack and conda recipes for fre-nctools, hsm, and the new fre-cli. Should they all be together (e.g. NOAA-GFDL/conda-recipes), or one recipe collection per package (e.g. NOAA-GFDL/fre-nctools-recipes)?

underwoo:
For now, I think a single repository for all the recipes.  For an example look at how [conda-forge](https://github.com/conda-forge) does it. While they do have separate repositories for the recipes, they combine them in a _feedstock_ repository. We can separate them later if the one repository becomes too unwieldy.

ceblanton:
Thanks, @underwoo. That sounds like a plan- one NOAA-GFDL repository for all conda recipes, like conda-forge. Can you create the repository, or is there a form for that? I think we would want it in the least surprising public place (https://github.com/NOAA-GFDL/conda-recipes). The individual recipes would be a subdirectory that could be a submodule as needed (e.g. recipes/fre-nctools).

Although, I wonder if we should raise the ante and make our recipes officially part of conda-forge. We can always do that later, I suppose. (conda-forge requires that all prerequisites be also in conda-forge, I think, so may be difficult for some packages).

mlee03:
@ceblanton, is this draft PR still in progress?  If not, can we close it?

ceblanton:
Thanks @mlee03.. we agreed to keep conda recipes for GFDL packges in a single place (hopefully associated with a build system like conda-forge) rather than with the code, so this is only notes and not something we want to merge. Let's close.

PR #233, opened 2023-06-26
GPU:  move non create_xgrid* functions to a utils file 

mlee03:
In this PR,
1.  All OpenACC directives have been removed to start fresh.
2. All non create_xgrid functions in `create_xgrid.c` have been moved to `create_xgrid_util.c`

PR #235, opened 2023-07-11
refactor list allocation/init and test for create_xgrid.c

rem1776:
Refactors the list allocation and min/max/avg calculations into their own function. Also moves the test file hidden at the bottom of create_xgrid.c into its own file + adds a script to run it.

this might need #233 first


mlee03 (approved):

rem1776:
closing this, the refactoring changes were already made to the dev branch and i don't think we need the test changes.

PR #246, opened 2023-09-13
introduce function to malloc arrays and introduce usage of realloc 

mlee03:
In this PR, 

- The function `malloc_arrays` is introduced to free and/or allocate the arrays _i_in, j_in, i_out, j_out, xgrid_clat, xgrid_clat, xgrid_area_.  This function will become handy when the algorithm is later updated to run on GPUs.  
- The function `realloc` is used to substitute the repeated malloc and free found in `setup_conserve_interp` where the computed values from create_xgrid are stored into the struct Interp


tools/fregrid/conserve_interp.c
```diff
@@ -92,304 +84,276 @@ void setup_conserve_interp(int ntiles_in, const Grid_config *grid_in, int ntiles
...
-	    free(tmp_i_out);
-	    free(tmp_j_out);
-	    free(tmp_area);
-	  }
-	}  /* if(nxgrid>0) */
-      }
-    }
+          }
+          else {
+            interp[n].i_in   = (int    *)realloc(interp[n].i_in,  interp[n].nxgrid*sizeof(int   ));
```
mlee03, line 204: I don't know why the spacing is so off.  Realloc is used here and all the tmp_* arrays have been removed

tools/fregrid/conserve_interp_util.c
```diff
@@ -105,11 +105,58 @@ void read_remap_file( int ntiles_in, int ntiles_out, Grid_config *grid_out,
...
+        free(xgrid_clat);
+      }
     }//if read from file
   } // ntiles
   if(mpp_pe() == mpp_root_pe())printf("NOTE: Finish reading index and weight for conservative interpolation from file.\n");
 
 }//end read_regrid_weights
+
+
+void malloc_arrays( int nsize, int **i_in, int **j_in, int **i_out, int **j_out,
```
ngs333, line 118: Please give a more meaningful name, like malloc_xgrid_arrays.

mlee03:
@ngs333 will make changes.  Btw, do you know another way to pass in an array/pointer to a function to allocate them in C? I don't like this whole double pointer thing.

tools/fregrid/conserve_interp_util.c
ngs333: @mlee03 There are lots of messes in the project, and some of them have a natural fix in other languages but not in C. This particular one is one of those.  First we don't have references in C, which would provide part of a natural solution to what you are trying to do. If we did instead of  vars in malloc_arrays() like "int **i_in" we would declare them as "int *& i_in" - a pointer reference so that one could modify the ultimate variable  (like in Fortran) . It would be easier to code the function and to use, and it would be safer. Then if you had templates and variable function args, you can code it like the one I have for C++ (below) - which works for any number and type of arguments (notice sometimes we don't use xgrid_clon and xgrid_clat and we need a second function,  but then function overloading is yet not allowed in C anyways). Its like this:

```
template <typename First>
void reaquire_memory(const size_t sz, First *& first){
  if (first != nullptr)
    free(first);
    first = (First *) malloc(sz * sizeof(First));
}
template <typename First, typename... Rest>
void reaquire_memory(const size_t sz, First*& first, Rest*&... rest){
  reaquire_memory( sz, first );
  reaquire_memory( sz, rest... );
}
```
And I call it like this:
 ```
reaquire_memory(get_MAXXGRID(), i_in, j_in, i_out, j_out, xgrid_area );
```
and like this
```  reaquire_memory(get_MAXXGRID(), i_in, j_in, i_out, j_out, xgrid_area, xgrid_clon, xgrid_clat);```

And while we are on it, the prior developers did not make a good choice by not putting all those fields ( ``` i_in, j_in, i_out, j_out, xgrid_area, xgrid_clon, xgrid_clat ``` ) in a structure, but then since they didn't have good unit tests frameworks (therefore good unit tests) it would be hard to change and guarantee that it works.

In any case, it was very smart of you to notice that this part of the code should be refactored, but you are working in C which has limited capabilities and on a project without a true unit test framework  Perhaps one day you might want to work on the C++ project where Everything is easier :-)

Thanks
MZ

ngs333 (approved):

PR #249, opened 2023-10-11
Create structures for list arrays and Add OpenACC data transfer directives 

mlee03:
In this PR,
- structure `Minmaxavglists` has been declared to hold all "list" arrays pertaining to the output grid
- function  `get_minmaxavg_lists` has been created.  This function computes values for the "list" arrays in structure Minmaxavglists
- `OpenACC` directives to transfer data to/from the device has been added.

### DIFF
--- a/tools/fregrid/globals.h
+++ b/tools/fregrid/globals.h
@@ -245,6 +245,6 @@ typedef struct{
   double *lon_list;
   double *lat_list;
   int *n_list;
-} Minmaxavglists;
+} Minmaxavg_lists;
 
 #endif
--- a/tools/fregrid/globals.h
+++ b/tools/fregrid/globals.h
@@ -96,7 +96,7 @@ typedef struct {
   char   area_name[STRING];
   int    do_regrid;
   int    is_axis_data;
-  int    dimsize[5];  
+  int    dimsize[5];
 } Var_config;
 
 typedef struct {
@@ -121,11 +121,11 @@ typedef struct {
   int  bndid;
   int  size;
   nc_type type;
-  char cart; 
+  char cart;
   int  bndtype;
   int  is_defined;
   double *bnddata;
-  double *data; 
+  double *data;
 } Axis_config;
 
 typedef struct {
@@ -235,5 +235,16 @@ typedef struct{
   double *f_max;
   double *f_min;
 } Monotone_config;
+
+typedef struct{
+  double *lat_min_list;
+  double *lat_max_list;
+  double *lon_min_list;
+  double *lon_max_list;
+  double *lon_avg;
+  double *lon_list;
+  double *lat_list;
+  int *n_list;
+} Minmaxavglists;
 
 #endif
--- a/tools/libfrencutils/create_xgrid.c
+++ b/tools/libfrencutils/create_xgrid.c
@@ -150,14 +150,14 @@ int create_xgrid_1dx2d_order2_(const int *nlon_in, const int *nlat_in, const int
 {
   int nxgrid;
   nxgrid = create_xgrid_1dx2d_order2(nlon_in, nlat_in, nlon_out, nlat_out, lon_in, lat_in, lon_out, lat_out, mask_in, i_in,
-                                     j_in, i_out, j_out, xgrid_area, xgrid_clon, xgrid_clat);
+             j_in, i_out, j_out, xgrid_area, xgrid_clon, xgrid_clat);
   return nxgrid;
 
 };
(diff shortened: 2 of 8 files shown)


tools/libfrencutils/create_xgrid_util.c
```diff
@@ -47,6 +47,15 @@ int line_intersect_2D_3D(double *a1, double *a2, double *q1, double *q2, double
 void malloc_minmaxavg_lists(const int n, Minmaxavglists *minmaxavglists)
 {
 
+  if( !minmaxavglists->lon_min_list ) free(minmaxavglists->lon_min_list);
```
ngs333, line 50: This may be backwards. Perhaps it should be like 
```if(  minmaxavglists->lon_min_list ) free(minmaxavglists->lon_min_list);```
I.e. the ```!``` should not be there? Or perhaps better to compare to NULL? Like:
```if( minmaxavglists->lon_min_list != NULL ) free(minmaxavglists->lon_min_list);``` ?

File: tools/fregrid/globals.h [hunk 2]
```diff
@@ -236,4 +236,15 @@ typedef struct{
...
+typedef struct{
+  double *lat_min_list;
+  double *lat_max_list;
+  double *lon_min_list;
+  double *lon_max_list;
+  double *lon_avg;
+  double *lon_list;
+  double *lat_list;
+  int *n_list;
+} Minmaxavglists;
```
ngs333, line 248: If you use a refactoring tool, it may be better to name this as
``Minmax_avg_lists``  ?

tools/libfrencutils/create_xgrid.c
```diff
@@ -212,19 +213,19 @@ int create_xgrid_1dx2d_order2(const int *nlon_in, const int *nlat_in, const int
...
-	  xgrid_clon[nxgrid] = poly_ctrlon(x_out, y_out, n_out, lon_in_avg);
-	  xgrid_clat[nxgrid] = poly_ctrlat (x_out, y_out, n_out );
-	  i_in[nxgrid]    = i1;
-	  j_in[nxgrid]    = j1;
-	  i_out[nxgrid]   = i2;
-	  j_out[nxgrid]   = j2;
-	  ++nxgrid;
-	  if(nxgrid > MAXXGRID) error_handler("nxgrid is greater than MAXXGRID, increase MAXXGRID");
-	}
+  if(xarea/min_area > AREA_RATIO_THRESH ) {
```
ngs333, line 218: Which editor or IDE are you using and is it it able to use the file FRE-NCTools/.editorconfig ? (PS Emacs and VSCode should be able to use it) I am not sure, but either :
a) your editor may be breaking the indentation of some of these lines. 
b) we may have to upgrade file .editorconfig
c) This file (and a few others) was never formatted post the intro of .editorconfig in the project. 
PS If you do decide to apply whole file reformatting, we may not be able to do so for the "usage" comment near 
the top.
PPS This also applies to other files (or functions) that were worked on.

ngs333 (changes requested):
I added a few comments. We might need a brief meeting to discuss some of them.

tools/libfrencutils/create_xgrid.c
mlee03: Why am I having these spacing issues all of a sudden?  Is this .editorconfig file new?

tools/libfrencutils/create_xgrid_util.c
mlee03: Made this more explicit

tools/libfrencutils/create_xgrid.c
mlee03: I did download the editorconfig plugin for emacs but these mysterious differences are still showing up

mlee03:
@ngs333, ready for merging with your approval.

tools/libfrencutils/create_xgrid.c
ngs333: @mlee03 
OK I'll accept it. Relating to this we've had a historical and informal modus operandi you might want to be aware of. With the intent of helping the reviewer, usually we have not reformatted entire files, but only those functions that were needed to be modified to fix the bug or add the enhancement. This ultimately stemmed from the fact that we have limited unit tests, and so reviewers must look carefully at those lines that are truly modified or changed; looking at other lines is extra work  and effort that detracts from checking correctness.  If it was desirable to reformat entire files just to upgrade the format, this was done in a separate push and possibly with a different PR.

ngs333 (approved):
Approving with two comments left elsewhere.

PR #251, opened 2023-10-19
create function to populate cell_in

mlee03:
This PR creates a new function called `get_CellStruct`.  This function populates the component arrays of cell_in.

### DIFF
--- a/tools/libfrencutils/mosaic_util.c
+++ b/tools/libfrencutils/mosaic_util.c
@@ -74,16 +74,6 @@ void error_handler(const char *msg)
   exit(1);
 #endif //ifdef use_libMPI
 
-#else
-
-  printf("Fatal Error: %s\n", msg);
-  //IMPERATIVE.  Must incorporate error handling
-  //#ifdef use_libMPI
-  //  MPI_Abort(MPI_COMM_WORLD, -1);
-  //#else
-  //  exit statements do not exist for OpenACC yet.
-  //#endif
-
 #endif //ifndef _OPENACC
 
 
--- a/tools/libfrencutils/mosaic_util.c
+++ b/tools/libfrencutils/mosaic_util.c
@@ -62,7 +62,6 @@ void set_rotate_poly_true(void){
     void error_handler(char *str)
     error handler: will print out error message and then abort
 ***********************************************************/
-#pragma acc routine seq
 void error_handler(const char *msg)
 {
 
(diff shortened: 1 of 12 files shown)


tools/libfrencutils/mosaic_util.c
```diff
@@ -62,7 +62,6 @@ void set_rotate_poly_true(void){
     void error_handler(char *str)
     error handler: will print out error message and then abort
 ***********************************************************/
-#pragma acc routine seq
 void error_handler(const char *msg)
```
ngs333, line 65: Can this function be cleaned up? 
Also, it seems that there is an "exit(1)" that can be executed but without printing anything. Is this true?

ngs333 (changes requested):
Please see comment.

mlee03:
I honestly have no idea what's happening, but I recently ran gpu_dev for benchmarking with respect to the CPU version and I'm seeing a lot more differences than I remember seeing several weeks ago (implying have I been doing something wrong this entire time?).  The differences are in `xgrid_area` with absolute max difference of _0.0197258_, and mean difference of _-3.06172e-05_ with stdev of _0.00179223_.

The other differences are seen with`tile1_distance` with absolute max difference of _5.77101e-11_, and mean difference of _-5.30129e-15_ with stdv of _ 8.7227e-13_

Are we good with these differences?

ngs333:
@mlee03 

> I honestly have no idea what's happening, but I recently ran gpu_dev for benchmarking with respect to the CPU version and I'm seeing a lot more differences than I remember seeing several weeks ago (implying have I been doing something wrong this entire time?). The differences are in `xgrid_area` with absolute max difference of _0.0197258_, and mean difference of _-3.06172e-05_ with stdev of _0.00179223_.
> 
> The other differences are seen with`tile1_distance` with absolute max difference of _5.77101e-11_, and mean difference of _-5.30129e-15_ with stdv of _8.7227e-13_
> 
> Are we good with these differences?

It depends. Max diff of 0.0197258 for an xgrid areas doesseem high; but if you know that the percent differences are sufficiently small, then its its OK. Note that cell areas can be ~10^7 meters square is *SOME* grids cells in some grids,  so if 0.0197258 meter square difference occurs for a large cell, then that is not really so bad.  What is the highest pct difference in the same test case?

mlee03:
The percentage difference for xgrid_area seems to be small, but seeing large percentage difference (up to 100%) for tile1_distance but these numbers are somewhere in the 10^-15

tools/libfrencutils/mosaic_util.c
mlee03: I haven't changed the exit(1) statement.
I think this is as clean as we can get this function right now. 
We need to look up what kind of error handling OpenACC has and I was going to leave that til the end.

ngs333:
> The percentage difference for xgrid_area seems to be small, but seeing large percentage difference (up to 100%) for tile1_distance but these numbers are somewhere in the 10^-15

This kind of error (tiny field values, large pct error) in floating point output is acceptable in CPU vs GPU comparisons.

PR #258, opened 2023-11-21
Fix for fregrid to properly sets NC4 compression

### LINKED ISSUE #257: fregrid fails with "Error in setting deflation level" when using Netcdf-4
**Describe the bug**
When using NetCDF-4 format input files without specifying the compression output level with `--deflation=`, fregrid sometimes fails with "Error in setting deflation level: NetCDF: Variable not found".

The problem is a bug within mpp_io.c that sets the compression level: https://github.com/NOAA-GFDL/FRE-NCtools/blob/master/tools/libfrencutils/mpp_io.c#L1024-L1072

The problem is that the variable ID is not consistent between the input and output file. Instead of assuming the variable ids in the input and output files are mapped to the same variables, the loop should be over the output variable IDs, and look up the input variable ID using the name.

**To Reproduce**
Using fregrid on any Netcdf-4 input file without specifying the compression level may reproduce the problem.

Using fregrid on a Netcdf input file with unregriddable variables (i.e. `interp_method=NONE`) while specifying all variables will definitely bring out the error.

**Expected behavior**
If compression settings are specified on the command line, then they should be used on the output file.

Otherwise, the compression settings of the input file should be used.

**System Environment**
Describe the system environment, include:
- OS: [e.g. RHEL 7.2]
- Compiler(s): Type and version [e.g. Intel 19.1]
- netCDF Version: For both C and Fortran
- Configure options: Any additional flags, or macros passed to configure

**Additional context**
Add any other context about the problem.  If applicable, include where any files
that help describe, or reproduce the problem exist.

ceblanton:
Also, comment-out the print statements as they are not useful to the user.

resolves #257

### DIFF
--- a/tools/libfrencutils/mpp_io.c
+++ b/tools/libfrencutils/mpp_io.c
@@ -29,6 +29,7 @@
 #define  MAXFILE 200
 #define  MAXVAR  1024
 #define  STRING 255
+#define  VARNAME_MAXSIZE 64
 
 typedef struct {
   int fldid;
@@ -1021,14 +1022,14 @@ void mpp_def_var_att_double(int fid, int vid, const char *attname, double attval
  * settings. If user doesn't specify (set to -1), the settings        *
  * of the input file are applied                                      *
  * ********************************************************************/
-void mpp_set_deflation(int fid_in, int fid_out, int deflation, int shuffle) {
-    // return if deflation set to zero
-    if (deflation == 0) {
-        printf("Not compressing due to option\n");
+void mpp_set_deflation(int fid_in, int fid_out, int user_deflation, int user_shuffle) {
+    // return if user is requesting no deflation
+    if (user_deflation == 0) {
+        //printf("Not compressing due to option\n");
         return;
     }
 
-    // return if netcdf3
+    // return if netcdf3 which does not support compression
     int format;
     char errmsg[512];
     int status;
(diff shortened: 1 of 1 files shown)


tools/libfrencutils/mpp_io.c
```diff
@@ -1037,37 +1038,53 @@ void mpp_set_deflation(int fid_in, int fid_out, int deflation, int shuffle) {
...
+    // loop thru output vars
+    int vid_in;
+    char varname[VARNAME_MAXSIZE];
+    int shuffle_in, deflate_in, deflation_in;
+    int shuffle_out, deflation_out;
+
+    for (int vid_out = 0; vid_out < files[fid_out].nvar; ++vid_out) {
+        // if user has not specified options, look up and use settings in input file
+        if (user_deflation == -1 || user_shuffle == -1) {
+            mpp_get_varname(fid_in, vid_out, varname);
```
ngs333, line 1056: Chris, this assumes that the variable_ids are 0,1,2,3,..., (files[fid_out].nvar -1) .  Is this true. I think this is usually true, but is it always true?

ceblanton: Thank you! The cargo cult programmer in me thought it would be true or maybe I saw it similar somewhere else and assumed (that's cargo cult also, ha ha!). Thanks for the review- all the in's and out's at least need a strong double-check and then there's things like this that I also took for granted. I agree with you `get_varid` has a similar use..

ngs333 (commented):
see comments

tools/libfrencutils/mpp_io.c
```diff
@@ -1037,37 +1038,53 @@ void mpp_set_deflation(int fid_in, int fid_out, int deflation, int shuffle) {
...
+            vid_in = mpp_get_varid(fid_in, varname);
+
+            status = nc_inq_var_deflate(files[fid_in].ncid, files[fid_in].var[vid_in].fldid, &shuffle_in, &deflate_in, &deflation_in);
```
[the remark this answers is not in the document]
ceblanton: Thanks so much for this. I added a comment distinguishing the MPP variable IDs (which are just a counter as you found) and the netCDF variable ID stored in the "fldid" structure, which is the true netCDF variable id (that should be used in netcdf method calls.)

ngs333 (approved):

ceblanton:
An example of the updates working properly could be done with:

```
tar -xf /archive/Raphael.Dussin/FMS2019.01.03_devgfdl_20230608/CM4_piControl_c192_OM4p125_v8/gfdl.ncrc5-intel22-prod-openmp/history/01800101.nc.tar ./01800101.ocean_annual.nc
tar -xf /archive/Raphael.Dussin/FMS2019.01.03_devgfdl_20230608/CM4_piControl_c192_OM4p125_v8/gfdl.ncrc5-intel22-prod-openmp/history/01800101.nc.tar ./01800101.ocean_static.nc
set vars = `list_ncvars.csh -st234 01800101.ocean_annual.nc | xargs | tr ' ' ,`
fregrid --standard_dimension --input_mosaic ocean_mosaic.nc --input_file 01800101.ocean_annual.nc --interp_method conserve_order1 --remap_file remap_file_native_to_1x1.nc --nlon 360 --nlat 180 --scalar_field $vars --output_file out.nc
```

PR #263, opened 2023-12-06
Include create_xgrid_util.h header file 

mlee03:
During the process of cleaning up FRE-NCTOOLS in gpu_dev, all miscellaneous non-create_xgrid* functions were moved to a new file called `create_xgrid_util.c` file. 

This change requires changes to all tools/*.c files that contains `#include "create_xgrid.h"`:  these c files also need to include `#include "create_xgrid_util.h"`

Without this change, answers produced with make_coupler_mosaic will disagree with those from the main branch.


mlee03:
@ngs333, c works in mysterious ways.  Just to double check, would you be able to test this branch to see if all agrees with the main branch?

tools/make_coupler_mosaic/Makefile.am
```diff
@@ -23,7 +23,8 @@ if WITH_MPI
 endif
 
 AM_CFLAGS = -I$(top_srcdir)/tools/libfrencutils \
-            $(NETCDF_CFLAGS)
+            $(NETCDF_CFLAGS) \
+            -I$(top_srcdir)/tools/fregrid
```
ngs333, line 27: I think other tools (make_coupler mosaic and the others) should not depend on the fregrid tool.  What is this tool picking up from the fregrid directory that the AM flags now have ```-I$(top_srcdir)/tools/fregrid``` ?

ngs333 (commented):
All looks fine except for one item. See commnent and question on make_coupler mosaic Makefile.am file.

tools/make_coupler_mosaic/Makefile.am
mlee03: The code does not compile without that include in AM_CFLAGS.  I think it's because create_xgrid.h needs global.h which is in the fregrid directory.

ngs333: @mlee03 So perhaps it makes more sense to put global.h in the same place as constant.h  - in directory libfrencutils ?

mlee03: Moved global.h to libfrenctools.  Ready for final review :)

ngs333 (approved):

PR #269, opened 2024-01-07
Optimizes function create_xgrid_great_circle to use the search tree.

ngs333:
This PR optimizes function **create_xgrid_great_circle** to use the search tree. I was optimized in a manner similar to create_xgrid_2dx2d_order{1|2}.  (Between them,  three of the four paths that **fregrid** uses  to create weights file, what remains to be optimized is the bilinear interpolation path). 

Also, we remove a few pieces of dead code that was intended for the AIX OS and which should not be
needed in C++.

As in the previous related enhancements, the changes were tested for exact reproduction of baseline results in creating weight files for mapping, including remapping  cubed sphere grid to RLL grid, RLL grid to RLL grid, and tripolar grids to RLL grids.
Mote that to test, one has to make grids (possibly through make_hgrid) with the option --great_circle_algorithm (otherwise the paths for conservative order1 or conservative order2 will be followed by fregrid).


cpp/fregrid/bilinear_interp.C
```diff
@@ -134,79 +144,84 @@ void setup_bilinear_interp(int ntiles_in, const Grid_config *grid_in, int ntiles
...
+  found = (int *) malloc(nx_out * ny_out * sizeof(int));
+  index = (int *) malloc(ntiles_in * 3 * sizeof(int));
+  shortest = (double *) malloc(ntiles_in * sizeof(double));
+  for (i = 0; i < nx_out * ny_out; i++) found[i] = 0;
+
+  dlon = dlon_in / nx_out;
+  dlat = dlat_in / (ny_out - 1);
+
+  for (iter = 1; iter <= max_iter; iter++) {
+    for (l = 0; l < ntiles_in; l++) {
```
ilaflott, line 156: Small nit-pick, but you did ask for a review, so... 

obviously one can tell `l` from `1` via syntax highlighting... but it's not exactly fun to have to rely on the syntax highlighter nor a few pixels of difference either.

But, i have a feeling `l` is used as a general index in multiple places, and so constitutes consistent style. It is also generally consistent with vector/tensor component notation. 

This isn't something that needs to be resolved here, but i want to comment this here now in case such a thing does become a headache for some reason in the future.

ilaflott:
This all looks well and fine to me. And particularly since it's a merge to a `dev` branch, as pointed out to me by rem1776, I'm inclined to simply approve.

ilaflott (approved):
I approve

rem1776 (approved):

PR #275, opened 2024-02-08
Added fix for blocking factor of 0 in mppnccombine

### LINKED ISSUE #272: mppnccombine data loss
Issue with `mppnccombine` reported by Martin Schmidt at the Leibniz-Institute for Baltic Sea Research.

When trying to use maximum block size with option `-k 0`, some time slices are lost if the
number of records in the file exceeds the maximum block size and is not
an integer multiple of `MAX_BF`. With all other choices of `-k` we get
correct results.

After line [1011](https://github.com/NOAA-GFDL/FRE-NCtools/blob/8176c2b45079431f2053ec773253938f8aee4828/postprocessing/mppnccombine/mppnccombine.c#L1011):

```c
       if ((*bf) >= 1)
         {
          if ((*bf) > (*nrecs)) {
            fprintf(stderr, "blocking factor (k) > total records (%d). Setting blocking factor to %d.\n",
                    *nrecs, *nrecs);
            *bf = *nrecs;
          }
          if (((*nrecs) % (*bf)) != 0) *nblocks = (int)((*nrecs)/(*bf)) + 1;
          else *nblocks = (int)((*nrecs)/(*bf));
         }
       else
         {
          /* bf was set to zero, so we do full buffering */
          *bf = min(MAX_BF,*nrecs); // we use the maximum blocking factor in our capacity
          /* normally we'll have one block, unless we hit MAX_BF */
          *nblocks = (int)((*nrecs)/(*bf));
         }
```

In the branch for `k=0` (`bf=0`), the case where "+ 1" is needed is missing, which results in the described loss of time slices.

The corrected code should be:
```c
       /* adjust bf */
       if ((*bf) >= 1)
         {
          if ((*bf) > (*nrecs)) {
            fprintf(stderr, "blocking factor (k) > total records (%d). Setting blocking factor to %d.\n",
                    *nrecs, *nrecs);
            *bf = *nrecs;
          }
          }
       else
         {
          /* bf was set to zero, so we do full buffering */
          *bf = min(MAX_BF,*nrecs); // we use the maximum blocking factor in our capacity
          /* normally we'll have one block, unless we hit MAX_BF */
         }
       /* find nblocks */
       if (((*nrecs) % (*bf)) != 0) *nblocks = (int)((*nrecs)/(*bf)) + 1;
       else *nblocks = (int)((*nrecs)/(*bf));
```

0cole:
Implemented the working fix in mppnccombine.c. A bug occurred when the blocking factor (`bf`) was set to 0 with the flag `-k 0`. In this case, if the number of records (`nrecs`) exceeded the maximum block size (`MAX_BF`) and the number of records was not an integer multiple of `MAX_BF`, time slices would have been lost.

The fix addresses this issue by ensuring that the correct number of blocks (`nblocks`) is calculated even when (`bf`) is set to 0. A subtest has been created in `Test02-mppnccombine.sh` that tests for this edge case where `bf` is set to 0.

Fixes #272


postprocessing/mppnccombine/mppnccombine.c
```diff
@@ -1004,29 +1004,28 @@ int process_vars(struct fileinfo *ncinfile, struct fileinfo *ncoutfile,
...
-        }
-        if (verbose) fprintf(stderr, "blocking factor=%d, num. blocks=%d, num. records=%d\n",*bf,*nblocks, *nrecs);
+   if (*nrecs==1) {
+     *nrecs=recdimsize;
+
+     /* adjust bf */
+     if ((*bf) >= 1) {
+       if ((*bf) > (*nrecs)) {
+         fprintf(stderr, "blocking factor (k) > total records (%d). Setting blocking factor to %d.\n",
+         *nrecs, *nrecs);
```
rem1776, line 1014: These args should have an indent since its part of the previous line

rem1776 (changes requested):
looks good! just one little nitpick for the code style

postprocessing/mppnccombine/mppnccombine.c
0cole: Just pushed the fix for this

rem1776 (approved):

PR #279, opened 2024-03-27
Move globals.h file from fregrid to libfrenctools directory

mlee03:
The `globals.h` header file contains all macro and struct definitions that can be used universally among all tools in FRE-NCTools.  Thus in this PR,  `globals.h`  has been moved  to the `libfrencutils` directory to better reflect the file's purpose.

Related to Issue #280


ceblanton (approved):

rem1776 (commented):
I think this file is also renamed in the other pr. Is there a reason we wanted to do this separately?

If not we can probably close this one and just use the other PR. I think it makes more sense to have the filename change alongside the changes done to the build files.

mlee03:
PR #282 depends on this PR.  I opened it as a separate PR in case there were any objections to  relocating the file.  Assuming there are no objections, I'll close this PR :)

PR #282, opened 2024-03-29
#1 Add OpenACC compile option to autotools 

mlee03:
In this PR, 

1.  _fregrid_acc_ and _libfrencutils_acc_ directories have been created.  The names of the files and the functions within have been changed with the suffix `_acc`.  All OpenACC development for OpenACC enabled `fregrid` will occur in these two directories.
2. _configure.ac_ and _Makefile.am_ have been modified to append `-acc` compiler flag to `FCFLAGS` and to compile `fregrid_acc`


rem1776 (commented):
I'm a bit confused by the copied over libfrencutils stuff. Is libfrencutils_acc a full copy of all the routines in the original libfrencutils with any gpu modifications needed? or is it supposed to only contain routines that were modified, with fregrid_acc using original versions when applicable?

I don't see fregrid_acc calling anything with the `_acc` added, so is it just using the orignal libfrencutils?

mlee03:
@rem1776 currently in this PR, all files  in the directories `fregrid_acc` and `libfrencutils_acc` are copies of those in `fregrid` and `libfrencutils` directories.  The only difference is that the functions in the *_acc files have the suffix "_acc".   
The compiled executables`fregrid` and `fregrid_acc` at this point should be identical. The OpenACC changes will be merged in in a stepwise fashion to make the reviewing process and backtracing easier.

rem1776:
> @rem1776 currently in this PR, all files in the directories `fregrid_acc` and `libfrencutils_acc` are copies of those in `fregrid` and `libfrencutils` directories. The only difference is that the functions in the *_acc files have the suffix "_acc". The compiled executables`fregrid` and `fregrid_acc` at this point should be identical. The OpenACC changes will be merged in in a stepwise fashion to make the reviewing process and backtracing easier.

So are you going to be removing the unused/unmodified routines in libfrencutils_acc as part of those updates?

If thats the case then i think we should probably only add those routines to begin with, maybe we can just add the acc versions to the same libfrencutils files instead of creating a copy? Saves us the confusion of having copies of things, and having to add a lot of code just to delete it since it'll all still be in the git history.

mlee03:
Got it got it, yup I was going to remove functions as we go along (also thought it would be easier for debugging), but it is needless code that has to be carried around and removed functions that are needed in the future can be added in.  Okay, will remove all repetitive functions in fregrid_acc and libfrencutils_acc

mlee03:
Perhaps it's ready.  Decided to start almost from scratch.

rem1776 (dismissed):
Looks good! thanks for making those changes i hope they weren't too annoying.

After this goes in I have an m4 file in the gpu_dev branch for the configure checks with the openacc flag, I'll make a PR for it. It'll just adjust the flag for different compilers if needed and make sure it works.

configure.ac
```diff
@@ -85,12 +85,17 @@ AS_IF([test x"$with_mpi" = x"yes"],
 AM_CONDITIONAL([WITH_MPI], [test x"$build_mpi" = x"yes"])
 AM_CONDITIONAL([WITH_MPI_TESTS], [test x"$build_mpi" = x"yes" -a -z "$SKIP_MPI_TEST"])
 
+#Build with OpenACC suport.  Default is 'no'
+AC_ARG_ENABLE([acc],
+  [AS_HELP_STRING([--enable-acc], [Build with OpenACC support.  Defaults to 'no'])] )
+AS_IF([test ${enable_acc:no} = "yes"], [enable_acc=yes], [enable_acc=no])
+AM_CONDITIONAL([ENABLE_ACC], [test $enable_acc = "yes"])
+
```
bensonr, line 93: When enable-acc is set to yes (true), is there a second test to ensure you are using nvc/nvfortran and an appropriate error message?  If not, is the plan to simply fail the check_headers for `openacc.h`?

mlee03: I haven't even thought of that, putting in a second check is needed here since users could be using gcc which openacc.h exists but -acc is the wrong flag. @rem1776 will your PR have this fix?  If not, I'll put in the fix here :)

mlee03: Ignore above comment, -acc flag has nothing to do with enable-acc here, but since we're only supporting nvhpc compilers for fregird_acc so far, that check should be put in.

configure.ac
```diff
@@ -117,6 +122,12 @@ AC_CHECK_PROGS(PROVE, [prove])
 
 AM_CONDITIONAL([WITH_CHECK_PROGS], [test -n "$PROVE" -a -n "$BATS"])
 
+# Check if openacc.h exists
+if test $enable_acc = yes ; then
+   AC_CHECK_HEADERS([openacc.h], [], [AC_MSG_ERROR(["Cannot find OpenACC header file"])] )
+   AC_MSG_WARN(Currently only NVIDIA compilers are supported to compile with OpenACC in FRE-NCTOOLS)
```
mlee03, line 128: @rem1776 @bensonr warning has been added to configure about compiler support

README.md
```diff
@@ -165,6 +165,11 @@ make install
 The option  `--with-mpi` to the `configure` command will configure for building
 parallel running versions of certain FRE-NCtools applications.
 
+### OpenACC-enabled NCTools applications
```
mlee03, line 168: @rem1776 @bensonr, comments are added also in the README

rem1776 (approved):

Issue #277, opened 2024-02-15
fregrid mpp_get_varname error when regridding Netcdf-4 history files


ceblanton:
**Describe the bug**
When attempting to regrid variables that are in a Netcdf-4 input file, fregrid fails with errors like:

Error from pe 0: mpp_io(mpp_get_varname): error in get varname from file 09460101.atmos_tracer.tile1.nc: NetCDF: Variable not found

The problem only occurs with Netcdf-4 files and a workaround is setting the deflation and shuffle level manually, e.g. `--deflation=0 --shuffle=0`

**To Reproduce**
Standard fregrid usage:

fregrid --standard_dimension --input_mosaic C192_mosaic.nc --input_file 09460101.atmos_tracer --interp_method conserve_order2 --remap_file .fregrid_remap_file_576_by_360.nc --nlon 576 --nlat 360 --scalar_field bk,pk,ps,sphum,temp,SOA,dust1,dust2,dust3,dust4,dust5,ssalt1,ssalt2,ssalt3,ssalt4,ssalt5,omphil,omphob,bcphil,bcphob,radon,climatology_bounds --output_file out.nc

**Expected behavior**
fregrid should regrid the variables if they are in the file

**System Environment**
Describe the system environment, include:
- OS: [e.g. RHEL 7.2]
- Compiler(s): Type and version [e.g. Intel 19.1]
- netCDF Version: For both C and Fortran
- Configure options: Any additional flags, or macros passed to configure

**Additional context**
Add any other context about the problem.  If applicable, include where any files
that help describe, or reproduce the problem exist.

mlee03:
@ceblanton, great timing!  It works!

uramirez8707:
After merging https://github.com/NOAA-GFDL/FRE-NCtools/pull/278
In order to reproduce the problem, the average_* variables need to be removed. For example:
```
netcdf wut2 {
dimensions:
        time = UNLIMITED ; // (2 currently)
        yh = 320 ;
        xh = 360 ;
        nv = 2 ;
variables:
        float SST(time, yh, xh) ;
                SST:_FillValue = 1.e+20f ;
                SST:missing_value = 1.e+20f ;
                SST:units = "degC" ;
                SST:long_name = "Sea Surface Temperature" ;
                SST:cell_methods = "area:mean yh:mean xh:mean time: mean" ;
                SST:cell_measures = "area: area_t" ;
        double time(time) ;
[...]

PR #285, opened 2024-04-16
Fix issue when getting the deflate lelvel of the input netcdf file

uramirez8707:
See https://github.com/NOAA-GFDL/FRE-NCtools/issues/277#issuecomment-2059388546 for details

### DIFF
--- a/tools/libfrencutils/mpp_io.c
+++ b/tools/libfrencutils/mpp_io.c
@@ -1050,7 +1050,12 @@ void mpp_set_deflation(int fid_in, int fid_out, int user_deflation, int user_shu
     for (int vid_out = 0; vid_out < files[fid_out].nvar; ++vid_out) {
         // if user has not specified options, look up and use settings in input file
         if (user_deflation == -1 || user_shuffle == -1) {
-            mpp_get_varname(fid_in, vid_out, varname);
+            mpp_get_varname(fid_out, vid_out, varname);
+
+            // Don't attempt to the get the deflate level if the variable is not in the input file (i.e lat_bnds,
+            // lon_bnds)
+            if (mpp_var_exist(fid_in, varname) == 0) continue;
+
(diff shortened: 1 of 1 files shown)


tools/libfrencutils/mpp_io.c
```diff
@@ -1050,7 +1050,12 @@ void mpp_set_deflation(int fid_in, int fid_out, int user_deflation, int user_shu
     for (int vid_out = 0; vid_out < files[fid_out].nvar; ++vid_out) {
         // if user has not specified options, look up and use settings in input file
         if (user_deflation == -1 || user_shuffle == -1) {
-            mpp_get_varname(fid_in, vid_out, varname);
+            mpp_get_varname(fid_out, vid_out, varname);
+
+            // Don't attempt to the get the deflate level if the variable is not in the input file (i.e lat_bnds,
+            // lon_bnds)
+            if (mpp_var_exist(fid_in, varname) == 0) continue;
+
```
ceblanton, line 1058: If the variable is not in the input file (e.g. lat_bnds) and the user doesn't specify the output settings, which output settings are used?

uramirez8707: In the case, there wouldn't be any compression for that variable:
```
        double lon_bnds(lon, bnds) ;
                lon_bnds:long_name = "longitude bounds" ;
                lon_bnds:units = "degrees_E" ;
                lon_bnds:axis = "X" ;
                lon_bnds:_Storage = "contiguous" ;
                lon_bnds:_Endianness = "little" ;
```
but the variable is nlonx2, so compressing it will not get you very far.

ceblanton: Thanks! sounds good to me

ceblanton (approved):

rem1776 (approved):

PR #287, opened 2024-05-09
Update configure.ac to reflect release 2024.01

ceblanton:


ceblanton:
@rem1776 I tagged 2024.01 but forgot to update configure.ac so will need to re-tag it.

Also, can this updating of configure.ac be automated somehow? I'm doomed to forgot this each time.. Thanks!

rem1776:
> @rem1776 I tagged 2024.01 but forgot to update configure.ac so will need to re-tag it.
> 
> Also, can this updating of configure.ac be automated somehow? I'm doomed to forgot this each time.. Thanks!

We could definitely have an action make a pull request with the version change, the main problem would be deciding how/what would trigger it.

If it was triggered by creating a tag then it wouldn't be included in that tag(unless it was retagged, which would trigger it again). We might be able to get around that though by using pattern matching with the tag name (like when we tag initially we append something to what should be the tag name).

The other option is doing what github calls a "workflow_dispatch", which would allow you to run the action manually and give it a tag name.

rem1776 (approved):

PR #289, opened 2024-05-13
3 - fregrid_acc: cleanup create_xgrid_acc

mlee03:
In this PR,

1.  create_xgrid functions that will not be supported (as of yet) with OpenACC will be removed.  These functions are those that, based on their name, create exchange grid between grids of differing dimensions (1d to 2d, 2d to 1d).  These have been removed under the assumption they are used infrequently and hence are of low priority.

2.  CPU OMP related code have been removed in the remaining create_xgrid functions.

3.  create_xgrid_utils_acc.c has been created and get_grid_area functions have been relocated to this newly created file.


mlee03:
closing, a larger PR will contain this PR.

PR #291, opened 2024-05-19
Adjust the size of MAXXGRID for serial/non-mpi builds

underwoo:
Other minor fixes.  Include:


ilaflott:
at gfdl, tested this with `--with-mpi` and non-mpi builds, using CM4 1/4 deg ocean model output. tested both builds with/without `--nthreads 16` argument. re-gridded file generated successfully in each case.

ilaflott (approved):

mlee03 (approved):

PR #292, opened 2024-05-23
Additional option "on_grid" for make_topog

ThomasNeumann2:
We introduced option "on_grid" which skips the regridding function. Using make_topog , we experienced changes in the topography even if the topography is on the target grid and "--dont_adjust_topo" is active. The reason seems to be that the topography is always passed through the regridding.

### DIFF
--- a/tools/make_topog/topog.c
+++ b/tools/make_topog/topog.c
@@ -358,7 +358,8 @@ void create_realistic_topog(int nx_dst, int ny_dst, const double *x_dst, const d
 			    int deepen_shallow, int full_cell, int flat_bottom, int adjust_topo,
 			    int fill_isolated_cells, int dont_change_landmask, int kmt_min, double min_thickness,
 			    int open_very_this_cell, double fraction_full_cell, double *depth,
-			    int *num_levels, domain2D domain, int debug, int use_great_circle_algorithm )
+			    int *num_levels, domain2D domain, int debug, int use_great_circle_algorithm,
+                int on_grid )
 {
   char xname[128], yname[128];
   int nx_src, ny_src, nxp_src, nyp_src, i, j, count, n;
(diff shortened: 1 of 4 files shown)


nikizadehgfdl:
@ThomasNeumann2,  I am curious to know the use-case  for this change. Could you provide and exmaple of where and how this mod is used? Thanks.

tools/make_topog/topog.c
```diff
@@ -497,13 +498,24 @@ void create_realistic_topog(int nx_dst, int ny_dst, const double *x_dst, const d
 	mask_src[i] = 1.0;
     }
   }
+  if(on_grid) {
```
mlee03, line 501: @ThomasNeumann2, looks good though I think lines 402 to here could be replaced by something slightly simpler.  The relevant info that comes out of those lines are the source grid indices that indicate where it overlaps with the dest grid.  Since the two grids are the same, indices from the domain struct could probably be used.

mlee03 (approved):

tools/make_topog/topog.c
ThomasNeumann2: @nikizadehgfdl : Our use case is the Baltic Sea. The models have a hor. grid resolution between 1n.m. and 3n.m. The model topography of the Baltic requires some manual modifications, small islands, channels, coast line, sills etc. A new topog run removes e.g. some islands. We found that the regridding function is doing it. Our solution is to skip this function if appropriate and maybe someone else could also made use of it.
@mlee03 : That is probably true. For me it was simpler to add the on_grid case and don't touch any other code. Thus, everything remains the same when using make_topog.

nikizadehgfdl (approved):

bensonr (approved):

rem1776 (approved):

PR #293, opened 2024-05-24
5 - fregrid_acc: rename variables, get_bounding_indices, get_input_skip_cells

mlee03:
In this PR,

1.  Many of the variables were renamed to better reflect their usage.  Of these variables, `interp` was renamed to `xgrid` to better indicate that the struct contains  exchange grid information.

2.  Function `get_bounding_indices` was defined to compute the lower and top input cell latitudinal indices.  These indices are the bounding points of the overlapping grid region.

3.  Function `get_skip_cells` was defined to set up what was formerly known as the "`mask`".  The "`mask`" is used to skip input cells in exchange grid creation and is not used in fregrid_acc (but it's used in FMS horiz_interp?)


rem1776:
Closing this based on the last comment

PR #278, opened 2024-03-25
Remove the use of average_* variables and use time_bnds variables instead

uramirez8707:
Fixes #273
Fixes #274

I tried this using output from the old diag manager (which includes the average_* variables) and this updated reproduces the master branch.

I tried this using output from the new diag manager (which does not have the average_* variables). This update reproduces the output from the old diag manager.

PR #294, opened 2024-05-29
Do not write time_avg_info variable attribute in fregrid

uramirez8707:
After https://github.com/NOAA-GFDL/FRE-NCtools/pull/278, fregrid does not write the average_* variables, but the time_avg_info variable attribute still exists.

This causes issues with [split_ncvars.pl](http://split_ncvars.pl/) fails when trying to splitting the regridded output. It is seeing the "average_T1" in the time_avg_info attribute but fails when trying to include it in the split-out output.

With this PR, the time_avg_info attribute is not longer written.


ceblanton:
I think we want this in the impending 2024.02 release that is unfortunately already tagged. So I think we'll have to re-tag as 2024.03.

ceblanton:
I'm going to re-tag 2024.02 as it's so fresh and no one is using it

ceblanton (approved):

PR #298, opened 2024-06-04
Change split_ncvars.pl regex to work for dimensionless variables

### LINKED ISSUE #297: split_ncvars.pl problem with dimensionless variables
**Describe the bug**
FRE pp (frepp) uses split_ncvars.pl to create per-variable timeseries and using the "single-file" -f option appends static variables to single pp file. In the latter use case, frepp first determines the static variables using `list_ncvars.csh -s`, then loops over the list of variables, running this for each variable `var`:

   split_ncvars.pl -s -v var -f /path/to/output.nc input.nc

split_ncvars.pl checks that the variable is in the input file, and if not, it is skipped (line 175):

```
if ($dump !~ /\t\w+ $var\(.+\)/) {
    print "WARNING: variable $var does not exist ... skipping.\n"
    next;
```

That regex works if the variable has any dimensions; if it has no dimensions (unusual but valid), the regex does not work.

**To Reproduce**
Obtain an input file that has a variable with no dimensions. Then try to extract the one variable with:

split_ncvars.pl -s -v var -f output.nc

WARNING: variable p700 does not exist ... skipping.

**Expected behavior**
split_ncvars.pl should append the variable to the output file even if it has no dimensions.

**System Environment**
any

**Additional context**
Sample file that works. p700 has the degenerate scalar_axis dimensions so the regex works

/work/c2b/atmos_daily_tile1_olddm.nc

Similar file that does not work because p700 has no dimensions

/work/c2b/atmos_daily_tile1_newdm.nc

ceblanton:
The semicolon at the end of the each ncdump line can be used for the regex.

resolves #297

### DIFF
--- a/postprocessing/split_ncvars/split_ncvars.pl
+++ b/postprocessing/split_ncvars/split_ncvars.pl
@@ -172,7 +172,7 @@ foreach my $file (@ifiles) {
          $var =~ s/^\s+//; $var =~ s/\s+$//;
 
          # skip this variable if it does not exist
-         if ($dump !~ /\t\w+ $var\(.+\)/) {
+         if ($dump !~ /\t\w+ $var\s+;/) {
            print "WARNING: variable $var does not exist ... skipping.\n" if !$Opt{QUIET};
            next;
          }


ceblanton:
@uramirez8707, this gets past the problem we saw with the modern diag manager output.

There could be something strange with the history files too, but split_ncvars.pl should be able to handle dimension-less variables however they got there.

chanwilson (approved):

PR #301, opened 2024-06-20
Fix index error and incorrect vector input  in setup_bilinear_interp

mlee03:
The portion of the code in setup_bilinear where the entire input grid is scanned to find the nearest  input cells to the output cell location seem to contain two bugs.  

The incrementation of the for loop index is changed from `j` to `jc`
Great circle distance is fixed and is now computed between vectors v0 and `v1`, not v0 and `v2`.

This section of the code is executed only if the first round of finding "nearest" input cells to an output cell fails and may have gone unnoticed?


rem1776 (approved):

bensonr (approved):
Those are some serious typos!!

PR #304, opened 2024-07-17
Add sync calls before any netcdf files are closed and add error checking

rem1776:
- Adds calls to `nf/nc_sync` before any calls to `nf/nc_close` 
- Adds error checking for close and sync return statuses if it wasn't done already
- Changes the error checking macros to more clear names (plus I think leading/trailing underscores are usually reserved for macros that are automatically set)
 
I also had to rename some files from `.f90` to `.F90`, it needs the uppercase F for the preprocessor to run.

### DIFF
--- a/postprocessing/land_utils/is-compressed.c
+++ b/postprocessing/land_utils/is-compressed.c
@@ -22,7 +22,7 @@
 #include <unistd.h>
 #include <netcdf.h>
 
-#define __NC_ASRT__(ierr) check_error(ierr,__FILE__,__LINE__)
+#define CHECK_NC_ERRSTAT(ierr) check_error(ierr,__FILE__,__LINE__)
 
 void usage(const char* name)
 {
@@ -49,8 +49,8 @@ void usage(const char* name)
 void check_error(int ierr, const char* file, int line)
 {
    if (ierr!=NC_NOERR) {
-      fprintf(stderr,"ERROR :: %s\n",
-	      nc_strerror(ierr));
+      fprintf(stderr,"ERROR :: %s\nfile: %s, line: %d\n",
+	      nc_strerror(ierr), file, line);
       exit(ierr);
    }
 }
@@ -85,18 +85,19 @@ int main(int argc, char* argv[])
 
    int optind = argc - 1;
    
-   __NC_ASRT__(nc_open(argv[optind],NC_NOWRITE,&ncid));
-   __NC_ASRT__(nc_inq_ndims(ncid,&ndims));
+   CHECK_NC_ERRSTAT(nc_open(argv[optind],NC_NOWRITE,&ncid));
+   CHECK_NC_ERRSTAT(nc_inq_ndims(ncid,&ndims));
    if (verbose) printf("Found %d dimensions\n",ndims);
    for (dimid=0; dimid<ndims; dimid++) {
-      __NC_ASRT__(nc_inq_dimname(ncid,dimid,name));
+      CHECK_NC_ERRSTAT(nc_inq_dimname(ncid,dimid,name));
       if (verbose) printf("Examining \"%s\"\n",name);
       if (nc_inq_varid(ncid,name,&varid)==NC_NOERR &&
           nc_inq_attid(ncid,varid,"compress",&attid)==NC_NOERR) {
           return 0;
       }
    }
-   __NC_ASRT__(nc_close(ncid));
+   CHECK_NC_ERRSTAT(nc_sync(ncid));
+   CHECK_NC_ERRSTAT(nc_close(ncid));
 
    return -1;
 }
--- a/postprocessing/land_utils/scatter-ncc.F90
+++ b/postprocessing/land_utils/scatter-ncc.F90
@@ -25,7 +25,7 @@
 !
 !-----------------------------------------------------------------------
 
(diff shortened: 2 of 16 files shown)


postprocessing/land_utils/is-compressed.c
```diff
@@ -85,18 +85,19 @@ int main(int argc, char* argv[])
...
-      __NC_ASRT__(nc_inq_dimname(ncid,dimid,name));
+      CHECK_NC_ERRSTAT(nc_inq_dimname(ncid,dimid,name));
       if (verbose) printf("Examining \"%s\"\n",name);
       if (nc_inq_varid(ncid,name,&varid)==NC_NOERR &&
           nc_inq_attid(ncid,varid,"compress",&attid)==NC_NOERR) {
           return 0;
       }
    }
-   __NC_ASRT__(nc_close(ncid));
+   CHECK_NC_ERRSTAT(nc_sync(ncid));
```
bensonr, line 99: With the `NC_NOWRITE` option to the open, this application will not be modifying the file and shouldn't need to sync prior to closing.

postprocessing/land_utils/scatter-ncc.F90
```diff
@@ -289,26 +289,28 @@ program combine_res
...
                  enddo
               endif
            endif
         endif
         deallocate(buffer,mask)
      enddo
   enddo
-  __NF_ASRT__(nf_close(in_ncid))
+  CHECK_NF_ERRSTAT(nf_sync(in_ncid))
+  CHECK_NF_ERRSTAT(nf_close(in_ncid))
```
bensonr, line 309: The descriptor `in_ncid` is associated with a file opened with `NF_NOWRITE` and doesn't need to sync prior to the close.

tools/libfrencutils/read_mosaic.c
```diff
@@ -80,6 +80,11 @@ int field_exist(const char* file, const char *name)
   else
     existed = 0;
 
+  status = nc_sync(ncid);
+  if(status != NC_NOERR) {
+    sprintf(msg, "field_exist: in syncing file %s.", file);
+    handle_netcdf_error(msg, status);
+  }
```
bensonr, line 87: The file associated with `ncid` is opened with `NC_NOWRITE` and doesn't need to sync before closing.  I think you can remove this logic and all other instances in this file.

tools/mppncscatter/mppncscatter.c
```diff
@@ -1415,11 +1417,14 @@ int mppncscatter(mnsopts* opts)
     fflush(stdout);
   }
 
-  nc_close(nc);
+  CHECK_NC_ERRSTAT(nc_sync(nc));
+  CHECK_NC_ERRSTAT(nc_close(nc));
```
bensonr, line 1421: I can't find whether this is a dryrun or files have been updated and the sync is needed prior to closing.

tools/ncexists/ncexists.c
```diff
@@ -137,6 +137,8 @@ int main (int argc, char **argv)
 	 }
 
 
+   status = nc_sync(ncid);
+       if (status != NC_NOERR) handle_error(status);
```
bensonr, line 141: Are we actually manipulating the contents of the file in the ncexists application?  If we aren't, then all opens should be of the nowrite variety and there will be no need to sync prior to close.

tools/simple_hydrog/lakes/cr_lake_files.F90
```diff
@@ -268,7 +270,8 @@ program cr_lake_files
    enddo
 enddo
 
-rcode= nf_close (ncid)
+CHECK_NF_ERRSTAT(nf_sync(ncid))
+CHECK_NF_ERRSTAT(nf_close(ncid))
```
bensonr, line 274: It appears `ncid` is associated with a file that is `NF_NOWRITE` and the sync may not be needed here.  Check the instances below in this same source file.

bensonr (changes requested):
I checked all but the last two files (cp_river_vars.F90 and rmv_parallel_rivers.F90).  Please make sure the files there are being modified and require the sync prior to close.

tools/mppncscatter/mppncscatter.c
rem1776: I removed this one, `nc` gets set on line 1289 and its opened with `NC_NOWRITE`. The loop below just needed the sync.

tools/ncexists/ncexists.c
rem1776: No it's not doing any writing it uses NC_NOWRITE for the open call. I removed the sync call and the nc_open call was just passing in 0 so i updated it to use the macro name from netcdf as well.

rem1776:
@bensonr This should be ready now, any syncs that were after read only open's should be removed.

ceblanton (approved):

ceblanton:
Thanks @rem1776 !   `combine-ncc`, `scatter-ncc`, and the river tools, all were clearly not syncing before.

Out of curiosity, were you able to bring out the bad behavior (exiting 0 before syncing, that we think occurred on gaea when F5 had issues) by hand?

bensonr (dismissed):
Looks good.  Thanks for making these changes.

bensonr (approved):
Looks good.  Thanks for making these changes.

PR #305, opened 2024-07-18
add m4 check for openacc flags and fix for intel

rem1776:
adds an m4 for checking valid openacc flags with compilers that support it (currently only nvc and gcc).

Also adds a header that was needed for mppncscatter to compile with intel due to the sync updates (forgot function order matters for C, gcc doesn't seem to care though) .


mlee03 (approved):
yay, thank you for the gcc -fopenacc support

PR #306, opened 2024-07-31
Update CI image

rem1776:
updates the ci image to a rocky 9 image with spack installed dependencies. The dockerfile is [here.](https://github.com/NOAA-GFDL/HPC-ME/pull/8)

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -12,7 +12,7 @@ jobs:
         with_mpi: ['','--with-mpi']
         enable_quad_precision: ['', '--enable-quad-precision']
     container:
-      image: ghcr.io/rem1776/fre-nctools-ci-rocky-gnu:14.1.0
+      image: ghcr.io/noaa-gfdl/fre-nctools-ci-rocky-gnu:14.1.0
       env:
         MPI: ${{ matrix.with_mpi }}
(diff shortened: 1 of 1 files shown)


ceblanton (approved):
Looks fine, but I confess I never completely understood how the `fre-nctools-base` image was stored and retrieved, and it looks like you're moving to this reasonably-looking approach:

https://docs.github.com/en/packages/working-with-a-github-packages-registry/working-with-the-container-registry

Is the `ghcr.io/noaa-gfdl` image available now, or should we wait?

rem1776:
> Looks fine, but I confess I never completely understood how the `fre-nctools-base` image was stored and retrieved, and it looks like you're moving to this reasonably-looking approach:
> 
> https://docs.github.com/en/packages/working-with-a-github-packages-registry/working-with-the-container-registry
> 
> Is the `ghcr.io/noaa-gfdl` image available now, or should we wait?

Yeah this is basically just moving from dockerhub to github and also updating to use the latest dependencies + gcc. The registry switch up was because there was a whole thing with dockerhub cancelling all free organization accounts (which we have), but i think they went back on it since. It doesn't make much of a difference to be honest, all registries generally work the same at the end of the day.

The image is ready now, it pulled the new image when it ran the CI for this pull request.

ceblanton:
Thank you!

PR #308, opened 2024-09-06
fregrid_acc - convert 2d lon/lat_vertices arrays to 1d arrays 

mlee03:
In this PR, 

`lon_vertices` and `lat_vertices` members in the `Grid_cell_struct_config` struct have been changed from double pointers to single pointers and are now malloc-ed to be 1d arrays.  This change has been made after testing showed that computation with 1d arrays is faster than with 2d arrays on GPUs.


bensonr (approved):
lgtm

rem1776 (approved):

PR #309, opened 2024-09-11
update deprecated `upload-artifact` action version

rem1776:
Github just deprecated the version of the action that uploads the build log, so this just bumps its up to the most recent version.


ceblanton:
thanks!

PR #312, opened 2024-09-24
Let split_ncvars split large Netcdf-4 files

### LINKED ISSUE #311: split_ncvars.pl specifies NetCDF-3 output unnecessarily
**Describe the bug**
split_ncvars.pl fails to split high-resolution (1/12 degree) ocean tripolar files because it specifies Netcdf-3 (64-bit offset) output format, even if the input format is Netcdf-4.

**To Reproduce**
Attempt to split a large Netcdf-4 ocean file.

```
an103:/work/c2b/big-ocean%>split_ncvars.pl 19620101.ocean_z_month.nc -o split -V      
split_ncvars.pl version 20190813
/app/spack/2023.02/linux-rhel7-x86_64/gcc-12.3.0/nco/5.1.5-bqjws3c7utjz7dqkobcprt7xsbuj75rz/bin/ncks --fl_fmt=64bit_offset --header_pad 16384 -h  -v so,z_i,average_T1,average_T2,average_DT,time_bnds 19620101.ocean_z_month.nc .split_ncvars.35210.var.nc

ERROR NC_EVARSIZE One or more variable sizes violate format constraints
HINT: NC_EVARSIZE errors occur when attempting to copy or aggregate input files together into an output file that exceeds the per-file capacity of the output file format, and when trying to copy, aggregate, or define individual variables that exceed the per-variable constraints of the output file format. The per-file limit of all netCDF formats is not less than 8 EiB on modern computers, so any NC_EVARSIZE error is almost certainly due to violating a per-variable limit. Relevant limits: netCDF3 NETCDF_CLASSIC format limits fixed variables to sizes smaller than 2^31 B = 2 GiB ~ 2.1 GB, and record variables to that size per record. A single variable may exceed this limit if and only if it is the last defined variable. netCDF3 NETCDF_64BIT_OFFSET format limits fixed variables to sizes smaller than 2^32 B = 4 GiB ~ 4.2 GB, and record variables to that size per record. Any number of variables may reach, though not exceed, this size for fixed variables, or this size per record for record variables. The netCDF3 NETCDF_64BIT_DATA and netCDF4 NETCDF4 formats have no variable size limitations of real-world import. If any variable in your dataset exceeds these limits, alter the output file to a format capacious enough, either netCDF3 classic with 64-bit offsets (with -6 or --64), to PnetCDF/CDF5 with 64-bit data (with -5), or to netCDF4 (with -4 or -7). For more details, see http://nco.sf.net/nco.html#fl_fmt
```

**Expected behavior**
Split the Netcdf-4 input file into per-variable Netcdf-4 output files.

**System Environment**
any

**Additional context**
Add any other context about the problem.  If applicable, include where any files
that help describe, or reproduce the problem exist.

ceblanton:
resolves #311

### DIFF
--- a/postprocessing/split_ncvars/split_ncvars.pl
+++ b/postprocessing/split_ncvars/split_ncvars.pl
@@ -66,17 +66,14 @@ my $odir = $Opt{odir};
 
 my $ncrcat = `which ncrcat 2>&1`;
 die "Unable to locate ncrcat, cannot continue.\n" if $ncrcat =~ /(: no)|(not found)/;
-chomp $ncrcat; $ncrcat .= " --64 -t 2 --header_pad 16384";
+chomp $ncrcat; $ncrcat .= " -t 2 --header_pad 16384";
 
 my $ncks = `which ncks 2>&1`; 
 die "Unable to locate ncks, cannot continue.\n" if $ncks =~ /(: no)|(not found)/;
 chomp $ncks; 
 # NCO changed options between 4.5.4 and 4.5.5 and 4.6. Hopefully they stopped those schenanigans.
 my $tmp = qx{$ncks --version 2>&1};
-$ncks .=  ($tmp =~ /4\.5\.4/ ? ' --64bit'
-	   : $tmp =~ /4\.5\.5/ ? ' --64bit_offset'
-	   : ' --fl_fmt=64bit_offset')
-       . ' --header_pad 16384';
+$ncks .= ' --header_pad 16384';
 
 my $list_ncvars = `which list_ncvars.csh 2>&1`;
 die "Unable to locate list_ncvars.csh, cannot continue.\n" if $list_ncvars =~ /(: no)|(not found)/;
--- a/postprocessing/split_ncvars/split_ncvars.pl
+++ b/postprocessing/split_ncvars/split_ncvars.pl
@@ -71,8 +71,6 @@ chomp $ncrcat; $ncrcat .= " -t 2 --header_pad 65536";
 my $ncks = `which ncks 2>&1`; 
 die "Unable to locate ncks, cannot continue.\n" if $ncks =~ /(: no)|(not found)/;
 chomp $ncks; 
-# NCO changed options between 4.5.4 and 4.5.5 and 4.6. Hopefully they stopped those schenanigans.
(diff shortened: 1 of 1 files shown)


ilaflott:
Another thought- given that the desired behavior is `input_file_type == output_file_type`, if the input file is of the 64-bit offset variety [enabling larger file output](https://www.earthdata.nasa.gov/s3fs-public/imported/ESDS-RFC-011v2.00.pdf), then isn't removal of the `ncks` version checking logic re-introducing potential for failures when the version of `ncks` isn't quite what's expected? wouldn't we need that logic to preset the offset?

ceblanton:
Thanks Ian. Good questions. Your last question is the easiest. The NCO version checking is only needed for outputting the 64-bit NC3 format, so if we remove that option, the NCO version can be ignored. (That's a nice document about netcdf, there!)

I see your point about flexibility. If split-ncvars.pl were used for multiple purposes and some of the purposes were format dependent, then I would agree with an --format option. But FRE is the only known user of split-ncvars, and if there is an outcry we can add an option (But.. then we'd have to check the NCO version probably). Users can always use `nccopy` or even `ncks` to convert formats.

The main reason to not add an extra flag to get the desired format behavior is that FRE would then have to check the version of the input file, just to pass it to the split tool. That wouldn't be terrible but would be unnecessary, since the desired behavior should be to output the same format as the input. That is the convention used by other tools, such as mppnccombine, fregrid, and timavg.csh.

ilaflott (approved):
I see, using 64-bit offset DEMANDS netcdf3 format. 

Looking through the development history of the script, [it seems that netCDF3 with 64-bit offset was the initial choice](https://github.com/NOAA-GFDL/FRE-NCtools/commit/8f0042d1fb7b5070ac0f8ec46acf646c1ad602c8). Given that points to a commit converting a c-shell implementation to a perl one, and the commit before that adds the c-shell script in for the first time, the original intention of specifying 64-bit is probably lost to time. 

I'll speculate that 64-bit offset around a one or two decades ago was likely the most flexible choice back then. I'll say that justifies an update now.

If there is user demand, there shouldn't be any reason we can't pass a file-format argument through to `ncks`'s `fl_fmt` field for greater control of output.

PR #317, opened 2024-10-11
Use full path for several applications

underwoo:
Use the full path for `ncdump`, and the NCO tools that are found during the configure step.  This will remove the need to have several modules loaded.

The downfall is that users cannot easily move to newer versions of these tools (using modules), and if the tools are removed (the particular version is removed) then the applications will fail.


chanwilson (approved):
looks ok at first blush.

rem1776:
@underwoo @ceblanton I'll build a new CI image for these updates, seems like its failing on one of the added checks for the netcdf operators.

underwoo:
I'm going to close this as #322 will do this as well -- and is better tested.

PR #324, opened 2024-11-15
Update the GFDL environments for PP/AN and GFDL-WS to use SW Env 2024.02

ceblanton:
And switch the PP/AN config to use the OS 8 modulefiles.

All `make check` tests pass.


HansVahlenkamp (approved):
Looks good.

ceblanton:
Thanks Hans. I know we don't really need them now that you deploy the GFDL-side fre-nctools via Spack but it's still handy for testing.

PR #326, opened 2024-12-05
Update mppnccombine.c to V2.2.8

HansVahlenkamp:
If no netCDF format is specified (no -64 or -n4 options) for the  output file then use the netCDF format of the first input file; use netCDF3 64-bit offset format or netCDF4 classic format.

### DIFF
--- a/postprocessing/mppnccombine/mppnccombine.c
+++ b/postprocessing/mppnccombine/mppnccombine.c
@@ -25,9 +25,9 @@
                  (http://www.gfdl.noaa.gov/~vb/mpp_io.html) by V. Balaji.
 
   V2.2.8: Hans.Vahlenkamp@noaa.gov
-          If no netCDF format is specified (no -64 or -n4 options) for the
-          output file then use the netCDF format of the first input file;
-          use netCDF3 64-bit offset format or netCDF4 classic format.
+          If the netCDF format for the output file is not chosen with the -64
+          or -n4 options then automatically use the netCDF format of the first
+          input file; netCDF3 64-bit offset or netCDF4 classic model.
   V2.2.7: Hans.Vahlenkamp@noaa.gov
           Synchronize output file before closing and check for errors.
   V2.2.6: Seth Underwood <Seth.Underwood@noaa.gov>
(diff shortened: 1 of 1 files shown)


underwoo:
@HansVahlenkamp is this ready?  If so, please remove the "draft" status of this PR.

HansVahlenkamp:
> @HansVahlenkamp is this ready? If so, please remove the "draft" status of this PR.

It passes my testing.

postprocessing/mppnccombine/mppnccombine.c
```diff
@@ -24,6 +24,10 @@
                  programming interface "mpp_io_mod"
                  (http://www.gfdl.noaa.gov/~vb/mpp_io.html) by V. Balaji.
 
+  V2.2.8: Hans.Vahlenkamp@noaa.gov
+          If no netCDF format is specified (no -64 or -n4 options) for the
+          output file then use the netCDF format of the first input file;
+          use netCDF3 64-bit offset format or netCDF4 classic format.
```
mlee03, line 30: Can the comment be a bit more explicit and say "users have the choice to request..."?

mlee03 (changes requested):

postprocessing/mppnccombine/mppnccombine.c
HansVahlenkamp: Ok, I reworded the version comment and the usage message a bit.

underwoo (changes requested):
The update is not complete, and has an issue in that it will still output a "classic" netcdf file.

I have a test case ready (see https://github.com/underwoo/FRE-NCtools/blob/test.and.source.cleanup/tests/mppnccombine/input-output-kinds).  Please use this test case to update mppnccombine.

I also feel we should not limit the output to NetCDF 4 with Classic.  We should support the full NetCDF 4 file (with HDF4/5).

HansVahlenkamp:
> The update is not complete, and has an issue in that it will still output a "classic" netcdf file.
> 
> I have a test case ready (see https://github.com/underwoo/FRE-NCtools/blob/test.and.source.cleanup/tests/mppnccombine/input-output-kinds). Please use this test case to update mppnccombine.
> 
> I also feel we should not limit the output to NetCDF 4 with Classic. We should support the full NetCDF 4 file (with HDF4/5).

Well, netCDF4 classic model format is still based on HDF5.  The mppnccombine "-n4" option has always written out netCDF4 classic model format.  That's why I did the same thing when the -n4 option is not used and the input is any netCDF4 to write out netCDF4 classic model format.  Is that too restrictive?  Are any GFDL models writing out netCDF4 that is not classic model?

underwoo:
> Well, netCDF4 classic model format is still based on HDF5. The mppnccombine "-n4" option has always written out netCDF4 classic model format. That's why I did the same thing when the -n4 option is not used and the input is any netCDF4 to write out netCDF4 classic model format. Is that too restrictive? Are any GFDL codes writing out netCDF4 that is not classic model format?

Not yet, but some of the components are using HDF5 features in the NetCDF4 files.  Since we are making the updates to mppnccombine now, we should future proof the output file instead of waiting.

As mentioned in the previous comment, the way the update is written now there is a bug in the logic.  The test still had a "classic" netcdf file written out.

HansVahlenkamp:
> > Well, netCDF4 classic model format is still based on HDF5. The mppnccombine "-n4" option has always written out netCDF4 classic model format. That's why I did the same thing when the -n4 option is not used and the input is any netCDF4 to write out netCDF4 classic model format. Is that too restrictive? Are any GFDL codes writing out netCDF4 that is not classic model format?
> 
> Not yet, but some of the components are using HDF5 features in the NetCDF4 files. Since we are making the updates to mppnccombine now, we should future proof the output file instead of waiting.
> 
> As mentioned in the previous comment, the way the update is written now there is a bug in the logic. The test still had a "classic" netcdf file written out.

There was previously no test for the netCDF cdf5 format of an input file, but this is now included.  There is also updated code to allow any netCDF format from the input file to be used for the output file; except if the input file is netCDF classic then netCDF 64-bit offset will be used instead.  (The "-n4" option still specifies using netCDF4 classic model format.)

underwoo (approved):

mlee03 (approved):

ceblanton (approved):

PR #356, opened 2025-03-21
Fregrid Optimization: Device Pointers

hderry:
**Description**
- During exchange grid generation and remapping data, allocate temporary arrays on the device.
            - Allocate data memory on device using acc_malloc -Remove instances of `data create` –  data will only be housed on device, so there is no need to match creation on host.

-  Modify `present` clause, removing data allocated memory using acc_malloc

- Utilize deviceptr clause to declare that pointers in list refer to a device pointer that does not need to be allocated / moved between host and device.

**How Has This Been Tested?**

- Standard build and installation process 
- Performed benchmark tests for fregrid_gpu with various input grids 

**Checklist:**
[✓] My code follows the style guidelines of this project
[✓ ] I have performed a self-review of my own code
[ ] I have commented my code, particularly in hard-to-understand areas
[ ] I have made corresponding changes to the documentation
[✓] My changes generate no new warnings
[ ] Any dependent changes have been merged and published in downstream modules
[ ] New check tests, if applicable, are included
[x ] `make distcheck` passes
split_ncvars/ fails but all other checks pass or are skipped

### DIFF
--- a/tools/fregrid_gpu/conserve_interp_gpu.c
+++ b/tools/fregrid_gpu/conserve_interp_gpu.c
@@ -18,6 +18,7 @@
  * <http://www.gnu.org/licenses/>.
  **********************************************************************/
 #include <stdlib.h>
+#include <openacc.h>
 #include <stdio.h>
 #include <string.h>
 #include <netcdf.h>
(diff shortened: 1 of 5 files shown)


mlee03:
@hderry, do you know which test is failing?

PR #364, opened 2025-06-03
add workflow for automatic tag creation

rem1776:
**Description**
adds a workflow that grabs the version number from the `configure.ac` and creates a new tag for releases. It will be triggered whenever `configure.ac` is modified but will not overwrite previously created tags. The grep command will match both `xxx.xx.xx` and `xxx.xx` style version numbers. End result basically ensures each new version number committed to main will have a corresponding tag.

Since my other PR is triggered by a tag (#357) and creates the release draft, once these both go in we'd have a (almost) fully automated release process. A PR containing a change to the version number would be merged in to main, and then we would just have to wait for the actions to run and manually release the created draft.

This kind of thing has potential for script injections, so I used a fairly strict grep command and called github's api directly (rather than using normal git commands). Hopefully that will mitigate any dangers, but any opinions are appreciated.

**How Has This Been Tested?**
my forks main branch

**Checklist:**
- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation
- [x] My changes generate no new warnings
- [x] Any dependent changes have been merged and published in downstream modules
- [x] New check tests, if applicable, are included
- [x] `make distcheck` passes

### DIFF
--- a/.github/workflows/tag.yaml
+++ b/.github/workflows/tag.yaml
@@ -1,0 +1,27 @@
+name: Tag for release
+on:
+  push:
+    paths:
+      - configure.ac 
+jobs:
+  tag_release:
+    runs-on: ubuntu-latest
+    steps:
+    - name: Checkout code
+      uses: actions/checkout@v4
+    - name: Get tag name
+      id: get-tag
+      run: echo "TAGNAME=`grep -o -E '?[0-9]{4}\.[0-9]{2}(\.[0-9]{2})?' configure.ac`" >> "$GITHUB_OUTPUT"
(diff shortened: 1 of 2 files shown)


ceblanton (approved):
Nice!!

ceblanton:
@rem1776 Yes, we want both, the `-dev` workflow and the auto-tagging workflow as well, thanks!

mlee03 (approved):

PR #409, opened 2026-06-11
consolidate workflow files to display CI workflow graph under workflow summary page

ilaflott:
**Description**
collapse `main`, `main_expensive`, `distcheck` workflow files --> one `main` workflow file with the same exact task dependencies

Closes #408 

**How Has This Been Tested?**
see checks, specifically the [summary](https://github.com/NOAA-GFDL/FRE-NCtools/actions/runs/27349169506?pr=409)

**Checklist:**
- [ ] My code follows the style guidelines of this project
- [ ] I have performed a self-review of my own code
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [ ] My changes generate no new warnings
- [ ] Any dependent changes have been merged and published in downstream modules
- [ ] New check tests, if applicable, are included
- [ ] `make distcheck` passes

PR #406, opened 2026-06-10
Fix intermittent parallel build failure in lib/libnfu

underwoo:
Under make -j, make scheduled two independent jobs that both compiled
  libnfu/nfu.F90: the .F90.o rule (producing nfu.o for libnfu.a) and the
  .F90.$(FC_MODEXT) suffix rule (producing nfu_mod.mod for
  nfu_compress.o's prerequisite). Because AM_FCFLAGS uses $(FC_MODOUT)$(@D),
  both gfortran processes wrote to the same libnfu/nfu_mod.mod0 temp file
  and raced to rename it, causing the intermittent error:

    f951: Fatal Error: Cannot rename module file 'libnfu/nfu_mod.mod0'
    to 'libnfu/nfu_mod.mod': No such file or directory

  Fix by adding an explicit no-op rule in lib/libnfu/local.mk:

    libnfu/nfu_mod.$(FC_MODEXT): libnfu/nfu.$(OBJEXT)
        @:

  This overrides the .F90.$(FC_MODEXT) suffix rule so nfu.F90 is compiled
  only once. nfu_mod.mod is treated as a side effect of building nfu.o,
  and nfu_compress.o waits for it via the existing prerequisite.

  Also replace the duplicate clean-local rule in local.mk (which produced
  a make warning) with CLEANFILES += entries, and add CLEANFILES = in
  lib/Makefile.am to initialise the variable before local.mk uses +=.

  Fix a typo in tools/simple_hydrog/lakes/Makefile.am: the prerequisite
  horiz_intrep.$(OBJEXT) did not match any source file; corrected to
  horiz_interp.$(OBJEXT).

  Closes: #405

**Description**
Include a summary of the change and which issue is fixed. Please also include
relevant motivation and context. List any dependencies that are required for
this change.

Fixes # (issue)

**How Has This Been Tested?**
Please describe the tests that you ran to verify your changes. Please also note
any relevant details for your test configuration (e.g. compiler, OS).  Include
enough information so someone can reproduce your tests.

**Checklist:**
- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation
- [x] My changes generate no new warnings
- [x] Any dependent changes have been merged and published in downstream modules
- [x] New check tests, if applicable, are included
- [x] `make distcheck` passes

PR #410, opened 2026-06-11
parallel nfu building + CI clarity changes

### LINKED ISSUE #405: Intermittent parallel build failure: f951 cannot rename nfu_mod.mod0 under make -j
## Summary

Running `configure && make -j` intermittently fails with:

```
f951: Fatal Error: Cannot rename module file 'libnfu/nfu_mod.mod0' to 'libnfu/nfu_mod.mod': No such file or directory
```

The failure is non-deterministic — it depends on which parallel jobs are scheduled first.

## Root cause

`lib/Makefile.am` sets:

```makefile
AM_FCFLAGS = $(FC_MODOUT)$(@D) ...
```

Modules are written **directly to `$(@D)`** (e.g. `libnfu/`), not to a `.mods/` subdirectory.  With `make -j`, two independent jobs both compile `libnfu/nfu.F90` simultaneously:

1. **`.F90.o` job** — builds `libnfu/nfu.$(OBJEXT)` (required by `libnfu.a`) and writes `libnfu/nfu_mod.mod` as a **side effect**.
2. **`.F90.$(FC_MODEXT)` suffix-rule job** — builds `libnfu/nfu_mod.$(FC_MODEXT)` (required by `libnfu/nfu_compress.$(OBJEXT)`) by compiling **the same source again**.

Both `gfortran` processes simultaneously write the temp file `libnfu/nfu_mod.mod0` and race to `rename(mod0 → nfu_mod.mod)`.  The loser finds its `.mod0` already consumed and dies with the error above.

The underlying cause is that `lib/libnfu/local.mk` has:

```makefile
libnfu/nfu_compress.$(OBJEXT): libnfu/nfu_mod.$(FC_MODEXT)
```

but no explicit rule for `libnfu/nfu_mod.$(FC_MODEXT)`, so make falls through to the `.F90.$(FC_MODEXT)` suffix rule and triggers a second compilation of `nfu.F90` in parallel with the first.

## Fix

Add an explicit no-op rule to `lib/libnfu/local.mk` that overrides the suffix rule and declares `nfu_mod.$(FC_MODEXT)` a side effect of `nfu.$(OBJEXT)`:

```makefile
libnfu/nfu_mod.$(FC_MODEXT): libnfu/nfu.$(OBJEXT)
	@:
```

This serialises the build: `nfu.o` is compiled first (writing `nfu_mod.mod` as a side effect), then `nfu_mod.$(FC_MODEXT)` is satisfied by the no-op rule, then `nfu_compress.o` proceeds.  `nfu.F90` is compiled only once.

Two additional housekeeping fixes are included:

- **`lib/Makefile.am`**: add `CLEANFILES =` initialisation so that `local.mk` can use `CLEANFILES +=` to register `libnfu/nfu_mod.$(FC_MODEXT)` and `libnfu/nfu_compress_mod.$(FC_MODEXT)` for cleanup, replacing the duplicate `clean-local` rule that was causing a make warning.
- **`configure.ac`**: add `AC_PROG_LN_S` so that `$(LN_S)` is defined and the `install-exec-hook` symlink rule for `mppdisttool` works correctly.

## Also fixed

A separate typo was found in `tools/simple_hydrog/lakes/Makefile.am` line 53:

```makefile
# before
horiz_interp_mod.mod: ... horiz_intrep.$(OBJEXT)
# after
horiz_interp_mod.mod: ... horiz_interp.$(OBJEXT)
```

`horiz_intrep.F90` does not exist; the correct source is `horiz_interp.F90`.  This would cause a build failure under `WITHOUT_EXTERN_FMS` whenever make needed to explicitly build `horiz_interp_mod.mod` as a target.

## Verification

Ten consecutive clean `make -j8` runs of `lib/` all succeed after the fix; all failed intermittently before it.

### LINKED ISSUE #408: make CI jobs / workflow clearer
i always find the CI here slippery because the branch/fork label doesn't get associated to the next jobs triggered...
_Originally posted by @ilaflott in https://github.com/NOAA-GFDL/FRE-NCtools/issues/406#issuecomment-4680799445_

ilaflott:
**Description**
- Sum of PRs #406 and #409
- adds `FORCE_JAVASCRIPT_ACTIONS_TO_NODE24: true` for prep of node20 retirement on the 19th
- ~~workflow now prints out the remote+branch being used, ty @singhd789 for [these bits](https://github.com/NOAA-GFDL/fre-workflows/blob/553b7c13a3a17d43a0ccb816974d291542dac243/.github/workflows/test_cloud_runner.yml#L49-L59)~~ oh right, test cloud runner is on parallel works, different rules...
- updates actions for js node24 migration warnings

Closes #405 and #408

**How Has This Been Tested?**
see checks

**Checklist:**
- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [ ] My changes generate no new warnings
- [ ] Any dependent changes have been merged and published in downstream modules
- [ ] New check tests, if applicable, are included
- [ ] `make distcheck` passes


ilaflott:
i think that's it.

`sanity_check_file.md` only for verification that the right code is being configured- this file can/will be removed pre-merge

sanity_check_file.md
ilaflott, line 1: removing pre-merge, leaving for now to assist with review

ilaflott:
this is ready for a solid review

ilaflott:
still desired? cleaning out old things in my github inbox

IMO this is the best of the CI clarity combined with the parallel build fix

.github/workflows/distcheck.yml
```diff
@@ -1,42 +0,0 @@
-# Github actions CI file
```
rem1776, line 1: my first workflow finally retired 🥲

.github/workflows/main.yml
```diff
@@ -3,40 +3,141 @@
...
+      - name: Configure
+        run: |
+          ls
+          pwd
+          mkdir build && cd build
+          autoreconf -i ../configure.ac
+          ../configure $MPI $QUAD_P || cat config.log
+
+      - name: Build distribution with check
+        run: make DISTCHECK_CONFIGURE_FLAGS="$MPI $QUAD_P" -C build -j
```
rem1776:
```suggestion
        run: make distcheck  -C build -j DISTCHECK_CONFIGURE_FLAGS="$MPI $QUAD_P"
```
needs the `distcheck` target specified, otherwise this will just do a normal build

rem1776 (changes requested):

.github/workflows/distcheck.yml
ilaflott: i think of it more as upgrading previous excellent work!

rem1776 (dismissed):
LGTM!

ceblanton (dismissed):
Strongly agree with this. Thank you!

ceblanton (dismissed):

